Issue #2, opened 2014-10-13
Add support for end date not been before start date


NSElvis:

callumboddy:
created a PR for this one:
https://github.com/hyperoslo/Form/pull/237

zenangst:
Fixed by https://github.com/hyperoslo/Form/pull/237

PR #8, opened 2014-10-16
Feature/photo field

3lvis:
ios simulator screen shot oct 16 2014 12 11 32 pm

### DIFF
--- a/Source/Cells/Custom/HYPImageFormFieldCell.m
+++ b/Source/Cells/Custom/HYPImageFormFieldCell.m
@@ -111,6 +111,13 @@ static const CGFloat HYPImageFormFieldContainerWidth = 360.0f;
 {
     [super layoutSubviews];
 
+    self.contentView.frame = [self contentViewFrame];
+}
+
+#pragma mark - Private methods
+
+- (CGRect)contentViewFrame
+{
     CGFloat horizontalMargin = HYPImageFormFieldHorizontalMargin;
     CGFloat verticalMargin = HYPImageFormFieldTopMargin;
 
@@ -120,7 +127,7 @@ static const CGFloat HYPImageFormFieldContainerWidth = 360.0f;
     frame.size.width = CGRectGetWidth(self.frame) - (horizontalMargin * 2);
     frame.size.height = CGRectGetHeight(self.frame) - verticalMargin;
 
-    self.contentView.frame = frame;
+    return frame;
 }
 
 @end


RyanKim01:
@3lvis, Hey I have 2 questions.
1) How do I put photo field? I tried putting 
{
              "id":"image",
              "type":"image",
              "size":{
                "width":100,
                "height":1.7
              }
            }
and nothing happened. and it even crashed at case FORMFieldTypeCustom: abort();
Are you purposely putting abort for custom fields?

2) Is there a closure for photo-field? I am not interested in picking images from uiimagepickercontroller but I want to initialize third party camera to take a photo in the closure. 

Thank you always!

3lvis:
Hi @RyanKim01, 

I understand at the moment is not so clear. Please have a look at https://github.com/hyperoslo/Form/tree/master/CustomField

- First your custom field should be a subclass of FORMBaseFieldCell
- Second you have to register your field to Form like this: https://github.com/hyperoslo/Form/blob/master/CustomField/RootController.swift#L20-L32

Form doesn't handle notifying you about actions on your custom field, you have to do it on your own. In the CustomField example, we do it using the `CustomFieldDelegate`.

I hope this is clear enough :grinning:

RyanKim01:
@3lvis ,as always, thank you for your kind explanation!
I have further questions.

1) So I basically made a new project using "CustomField" folder in Demo project. And I see that 
{
                                    "id":"textye",
                                    "type":"textye",
                                    "size":{
                                    "width":100,
                                    "height":2
                                    }
                                    }
is what's making a custom field. I get this part. But how do you make a section where you can make a camera or photo upload section like you did at the begging of this issue? Is there a way to make that using JSON format or did you hardcode that in CustomField class?

2. For the form I'm trying to make, I need to insert sub-heading. Let's say there is a field called "Last Name" and a textfield below it. If I wanna put a line of description above "Last Name", would it be possible? I was reading sample codes but It wasn't clear enough for me to figure this out.

"Sub-heading"
"Last Name"
(Text Field)

Above format is what I eventually want. If you are not comfortable with #2 question since it is not directly related to this issue, I will open a new issue.

3lvis:
Hi @RyanKim01, yes, please make an issue detailing what are you trying to accomplish, a small drawing would help.

Have a nice day!

RyanKim01:
@3lvis , Will do. But If would be great if you can answer my first question which is related to custom photo field! Thanks!

3lvis:
@RyanKim01 the photo field is a normal custom field named image in the JSON, we don't have that field out of the box.

https://github.com/hyperoslo/Form/blob/master/Basic-ObjC/Resources/forms.json#L88-L96

RyanKim01:
@3lvis , yeah i see that. But it was crashing the app everytime when I have below code in my app:
 {
              "id":"image",
              "type":"image",
              "size":{
                "width":100,
                "height":1.7
              },
              "hidden":false
            }

I set hidden as false since I wanna show photo field. But it seems like it is always crashing at
 case FORMFieldTypeCustom: abort();  (line 184 in FORMDataSource.m)
I wonder this is normal behavior or not. I just could not make the form like the image below.
screen shot 2015-12-17 at 4 32 09 pm

3lvis:
@RyanKim01 It's probably crashing because you haven't followed steps 1 and 2 from my comment.

https://github.com/hyperoslo/Form/pull/8#issuecomment-165379281

>1.- Your custom field should be a subclass of FORMBaseFieldCell like this: ?https://github.com/hyperoslo/Form/blob/master/CustomField/CustomField.swift
>2.- Register your custom field to Form like this: 
https://github.com/hyperoslo/Form/blob/master/CustomField/RootController.swift#L20-L32

PR #139, opened 2015-01-05
Add additional target actions

zenangst:
- [x] Add support for enable actions
- [x] Add support for disable actions
- [x] Add support for update actions with values
- [x] Add support for default values in select fields
- [x] Run targets on field values when form becomes enabled
- [x] Add tests

### DIFF
--- a/Demo/HYPForms/HYPSampleCollectionViewController.m
+++ b/Demo/HYPForms/HYPSampleCollectionViewController.m
@@ -49,7 +49,7 @@ HYPFormsCollectionViewDataSourceDataSource, HYPFormsLayoutDataSource>
     _formsManager = [[HYPFormsManager alloc] initWithJSON:JSON
                                             initialValues:self.initialValues
                                          disabledFieldIDs:@[@"email"]
-                                                 disabled:YES];
+                                                 disabled:NO];
 
     return _formsManager;
 }
--- a/Demo/HYPForms/HYPSampleCollectionViewController.m
+++ b/Demo/HYPForms/HYPSampleCollectionViewController.m
@@ -48,7 +48,7 @@ HYPFormsCollectionViewDataSourceDataSource, HYPFormsLayoutDataSource>
 
     _formsManager = [[HYPFormsManager alloc] initWithJSON:JSON
                                             initialValues:self.initialValues
-                                         disabledFieldIDs:@[@"email"]
+                                         disabledFieldIDs:@[@"display_name"]
                                                  disabled:YES];
 
     return _formsManager;
--- a/Demo/HYPForms/HYPSampleCollectionViewController.m
+++ b/Demo/HYPForms/HYPSampleCollectionViewController.m
@@ -49,7 +49,7 @@ HYPFormsCollectionViewDataSourceDataSource, HYPFormsLayoutDataSource>
     _formsManager = [[HYPFormsManager alloc] initWithJSON:JSON
                                             initialValues:self.initialValues
                                          disabledFieldIDs:@[@"email"]
-                                                 disabled:NO];
+                                                 disabled:YES];
 
     return _formsManager;
 }
--- a/Demo/HYPForms/HYPSampleCollectionViewController.m
+++ b/Demo/HYPForms/HYPSampleCollectionViewController.m
@@ -49,7 +49,7 @@ HYPFormsCollectionViewDataSourceDataSource, HYPFormsLayoutDataSource>
     _formsManager = [[HYPFormsManager alloc] initWithJSON:JSON
                                             initialValues:self.initialValues
                                          disabledFieldIDs:@[@"display_name"]
-                                                 disabled:YES];
+                                                 disabled:NO];
 
     return _formsManager;
 }
@@ -79,6 +79,7 @@ HYPFormsCollectionViewDataSourceDataSource, HYPFormsLayoutDataSource>
                 }
             }];
         }
+        
     };
 
     return _dataSource;
--- a/Demo/HYPForms/HYPSampleCollectionViewController.m
+++ b/Demo/HYPForms/HYPSampleCollectionViewController.m
@@ -48,7 +48,7 @@ HYPFormsCollectionViewDataSourceDataSource, HYPFormsLayoutDataSource>
 
     _formsManager = [[HYPFormsManager alloc] initWithJSON:JSON
                                             initialValues:self.initialValues
-                                         disabledFieldIDs:@[@"display_name"]
+                                         disabledFieldIDs:@[@"email"]
                                                  disabled:YES];
 
     return _formsManager;
--- a/Demo/HYPForms/HYPSampleCollectionViewController.m
+++ b/Demo/HYPForms/HYPSampleCollectionViewController.m
@@ -230,7 +230,13 @@ HYPFormsCollectionViewDataSourceDataSource, HYPFormsLayoutDataSource>
 {
     [self.dataSource disable:sender.isOn];
 
-    HYPFormTarget *target = [HYPFormTarget hideFieldTargetWithID:@"image"];
+    HYPFormTarget *target;
+
+    if (sender.isOn) {
+        target = [HYPFormTarget hideFieldTargetWithID:@"image"];
+    } else {
+        target = [HYPFormTarget showFieldTargetWithID:@"image"];
+    }
 
     [self.dataSource processTargets:@[target]];
 }
--- a/Demo/HYPForms/HYPSampleCollectionViewController.m
+++ b/Demo/HYPForms/HYPSampleCollectionViewController.m
@@ -152,10 +152,7 @@ HYPFormsCollectionViewDataSourceDataSource, HYPFormsLayoutDataSource>
 
     [self.navigationController setToolbarHidden:NO animated:YES];
 
-    HYPFormTarget *target = [HYPFormTarget new];
-    target.targetID = @"employment-1";
-    target.typeString = @"section";
-    target.actionTypeString = @"hide";
+    HYPFormTarget *target = [HYPFormTarget hideSectionTargetWithID:@"employment-1"];
 
     [self.dataSource processTargets:@[target, [HYPFormTarget hideFieldTargetWithID:@"image"]]];
 }
@@ -233,15 +230,7 @@ HYPFormsCollectionViewDataSourceDataSource, HYPFormsLayoutDataSource>
 {
     [self.dataSource disable:sender.isOn];
 
-    HYPFormTarget *target = [HYPFormTarget new];
-    target.targetID = @"image";
-    target.typeString = @"field";
-
-    if (sender.isOn) {
-        target.actionTypeString = @"hide";
-    } else {
-        target.actionTypeString = @"show";
-    }
+    HYPFormTarget *target = [HYPFormTarget hideFieldTargetWithID:@"image"];
 
     [self.dataSource processTargets:@[target]];
 }
--- a/Demo/HYPForms/HYPSampleCollectionViewController.m
+++ b/Demo/HYPForms/HYPSampleCollectionViewController.m
@@ -79,7 +79,6 @@ HYPFormsCollectionViewDataSourceDataSource, HYPFormsLayoutDataSource>
                 }
             }];
         }
-
     };
 
     return _dataSource;
--- a/Demo/HYPForms/HYPSampleCollectionViewController.m
+++ b/Demo/HYPForms/HYPSampleCollectionViewController.m
@@ -49,7 +49,7 @@ HYPFormsCollectionViewDataSourceDataSource, HYPFormsLayoutDataSource>
     _formsManager = [[HYPFormsManager alloc] initWithJSON:JSON
                                             initialValues:self.initialValues
                                          disabledFieldIDs:@[@"display_name"]
-                                                 disabled:NO];
+                                                 disabled:YES];
 
     return _formsManager;
 }
@@ -79,7 +79,7 @@ HYPFormsCollectionViewDataSourceDataSource, HYPFormsLayoutDataSource>
                 }
             }];
         }
-        
+
     };
 
     return _dataSource;
--- a/Demo/HYPFormsTests/HYPFormTargetTests.m
+++ b/Demo/HYPFormsTests/HYPFormTargetTests.m
@@ -29,6 +29,26 @@
     XCTAssertEqualObjects(target.actionTypeString, @"hide");
     XCTAssertTrue(target.type == HYPFormTargetTypeSection);
     XCTAssertTrue(target.actionType == HYPFormTargetActionHide);
+
+    target = [[HYPFormTarget alloc] initWithDictionary:@{@"id": @"first_name",
+                                                         @"type": @"field",
+                                                         @"action": @"disable"}];
+    XCTAssertNotNil(target);
+    XCTAssertEqualObjects(target.targetID, @"first_name");
+    XCTAssertEqualObjects(target.typeString, @"field");
+    XCTAssertEqualObjects(target.actionTypeString, @"disable");
+    XCTAssertTrue(target.type == HYPFormTargetTypeField);
+
+    XCTAssertTrue(target.actionType == HYPFormTargetActionDisable);
+    target = [[HYPFormTarget alloc] initWithDictionary:@{@"id": @"last_name",
+                                                         @"type": @"field",
+                                                         @"action": @"enable"}];
+    XCTAssertNotNil(target);
+    XCTAssertEqualObjects(target.targetID, @"last_name");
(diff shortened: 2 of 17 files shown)


Demo/HYPForms/HYPSampleCollectionViewController.m
```diff
@@ -233,15 +230,7 @@ - (void)readOnly:(UISwitch *)sender
...
-    HYPFormTarget *target = [HYPFormTarget new];
-    target.targetID = @"image";
-    target.typeString = @"field";
-
-    if (sender.isOn) {
-        target.actionTypeString = @"hide";
-    } else {
-        target.actionTypeString = @"show";
-    }
+    HYPFormTarget *target = [HYPFormTarget hideFieldTargetWithID:@"image"];
```
attila-at-hyper: Are you sure it's always "hide"? In the earlier version it depended on sender.isOn, now it seems to be always hide.

zenangst: You are right ;)

Demo/HYPFormsTests/HYPFormsCollectionViewDataSourceTests.m
```diff
@@ -82,6 +82,50 @@ - (void)testIndexInForms
...
+{
+    HYPFormField *targetField = [self.manager fieldWithID:@"display_name" includingHiddenFields:YES];
+
+    XCTAssertFalse(targetField.disabled);
+
+    HYPFormTarget *disableTarget = [HYPFormTarget disableFieldTargetWithID:@"display_name"];
+
+    [self.dataSource processTarget:disableTarget];
+
+    [self.manager fieldWithID:@"display_name" includingHiddenFields:YES];
```
attila-at-hyper: I'm not very familiar with HYPForms, but why do you need this line? Does it do some kind of processing? If it's only about gaining a pointer to a field (like on line 87), then it seems to be unnecessary, not to mention it's return value isn't used. I might easily be wrong though... :)

Demo/HYPForms/HYPSampleCollectionViewController.m
zenangst: Should be fixed by https://github.com/hyperoslo/HYPForms/commit/2a318b83ba05cbf42cab092fc0f3d0f9a37d6255

Demo/HYPFormsTests/HYPFormsCollectionViewDataSourceTests.m
zenangst: Correct again, oversight on my part. Was pulling my hair yesterday about one of the tests, might have been something related to that.

Removed in https://github.com/hyperoslo/HYPForms/commit/8bf35857685e7143889fc33c08dc9125d6895a46

Source/HYPFormsCollectionViewDataSource.m
```diff
@@ -78,11 +79,7 @@ - (instancetype)initWithCollectionView:(UICollectionView *)collectionView
 
 - (NSMutableArray *)collapsedForms
 {
-    if (_collapsedForms) return _collapsedForms;
-
-    _collapsedForms = [NSMutableArray new];
-
-    return _collapsedForms;
+    return _collapsedForms = (_collapsedForms) ?: [NSMutableArray new];
```
attila-at-hyper: Hm, what is the result of ?: if _collapsedForms is not nil? Will it be _collapsedForms (then it's OK) or nil? Maybe it would make the code more readable if you defined a value between ? and :.

Source/HYPFormsCollectionViewDataSource.m
```diff
@@ -304,11 +303,27 @@ - (void)disable:(BOOL)disabled
     }
 
     for (NSString *fieldID in fields) {
-        BOOL shouldDisable = (![fieldID isEqualToString:@"blank"] && ![self.formsManager.disabledFieldsIDs containsObject:fieldID]);
+        HYPFormField *field = [fields valueForKey:fieldID];
+        BOOL shouldChangeState = (![self.formsManager.disabledFieldsIDs containsObject:fieldID]);
```
attila-at-hyper: Have you intentionally ignored the case when fieldID is "blank"?

Source/HYPFormsCollectionViewDataSource.m
```diff
@@ -304,11 +303,27 @@ - (void)disable:(BOOL)disabled
...
             field.disabled = disabled;
+
+            if (field.targets.count > 0) {
+                [self processTargets:field.targets];
+            } else if (field.type == HYPFormFieldTypeSelect) {
+                BOOL hasFieldValue = (field.fieldValue && [field.fieldValue isKindOfClass:[HYPFieldValue class]]);
+                if (hasFieldValue) {
+                    HYPFieldValue *fieldValue = (HYPFieldValue *)field.fieldValue;
+                    if (fieldValue.targets.count > 0) {
+                        dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(0.1 * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{
```
attila-at-hyper: Just for the sake of curiosity: why do you have the 0.1 sec delay? Is 0.1 sec always the right duration for a delay?

Source/HYPFormsCollectionViewDataSource.m
zenangst: We no longer use blank fields in the layout so this should no longer be necessary

Source/HYPFormsCollectionViewDataSource.m
```diff
@@ -451,6 +466,13 @@ - (void)fieldCell:(UICollectionViewCell *)fieldCell updatedWithField:(HYPFormFie
     }
 }
 
+- (void)fieldCell:(UICollectionViewCell *)fieldCell processTargets:(NSArray *)targets
+{
+    dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(0.2 * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{
```
attila-at-hyper: In a previous change you used 0.1 seconds for the delay, now you use 0.2 seconds. Is that intentional? Shouldn't it be the same all the time? Maybe defined by a constant?

Source/HYPFormsCollectionViewDataSource.m
zenangst: Had some issues with ```processTargets``` calls clashing with the UICollectionView so I delayed it to ensure that the first call had been processed. A very wonky an unsafe solution and I'm pretty confident that this can be removed.

Gonna look into it a bit and see if this still needed

Source/HYPFormsManager.m
```diff
@@ -556,69 +563,114 @@ - (NSArray *)updateTargets:(NSArray *)targets
 
         if (!field) continue;
 
-        NSArray *fieldIDs = [field.formula hyp_variables];
```
attila-at-hyper: Uh, the below part seems like a "major" refactoring which is pretty difficult to be reviewed on a line by line bases. I'm happy to review it for you, but I'd need to have a look at the entire file and compare it to the old version. Would you like me to spend time on this?

Source/HYPFormsManager.m
```diff
@@ -556,69 +563,114 @@ - (NSArray *)updateTargets:(NSArray *)targets
...
+    NSMutableArray *indexPaths = [NSMutableArray new];
+
+    for (HYPFormTarget *target in targets) {
+        if (target.type == HYPFormTargetTypeSection) continue;
+        if ([self.hiddenFieldsAndFieldIDsDictionary objectForKey:target.targetID]) continue;
+
+        [self fieldWithID:target.targetID includingHiddenFields:YES completion:^(HYPFormField *field, NSIndexPath *indexPath) {
+            if (field) {
+                field.disabled = !enabled;
+                if (indexPath) [indexPaths addObject:indexPath];
```
attila-at-hyper: Since you create the indexPaths array on line 657, check if it isn't nil feels like an overkill here.

Source/HYPFormsCollectionViewDataSource.m
zenangst: ```_collapsedForms``` either gets assigned the value of the conditional, it will fallback to creating a new mutable array.

 I think this is pretty clearcut :)

Source/HYPFormsCollectionViewDataSource.m
zenangst: Removed in https://github.com/hyperoslo/HYPForms/commit/df424953ed443391e9926109b4108e0e7962d0ab

Source/HYPFormsManager.m
attila-at-hyper: Damn, I've missed the "s", sorry... :)

Source/HYPFormsCollectionViewDataSource.m
attila-at-hyper: I see the intention, but I'm a bit unsure about the "true" outcome, in other words: is `(_collapsedForms) ?: [NSMutableArray new]` equivalent to `(_collapsedForms) ? (_collapsedForms) : [NSMutableArray new]` or `(_collapsedForms) ? nil : [NSMutableArray new]`

Source/HYPFormsCollectionViewDataSource.m
attila-at-hyper: OK, didn't know, sorry :)

Source/HYPFormsManager.m
attila-at-hyper: I'm afraid I can't squeeze it in today, sorry.

Demo/HYPFormsTests/HYPFormTargetTests.m
```diff
@@ -29,6 +29,26 @@ - (void)testInitWithDictionary
...
+    target = [[HYPFormTarget alloc] initWithDictionary:@{@"id": @"first_name",
+                                                         @"type": @"field",
+                                                         @"action": @"disable"}];
+    XCTAssertNotNil(target);
+    XCTAssertEqualObjects(target.targetID, @"first_name");
+    XCTAssertEqualObjects(target.typeString, @"field");
+    XCTAssertEqualObjects(target.actionTypeString, @"disable");
+    XCTAssertTrue(target.type == HYPFormTargetTypeField);
+
+    XCTAssertTrue(target.actionType == HYPFormTargetActionDisable);
```
NSElvis: This should go with the block at the top, just after line 40

zenangst: Fixed in https://github.com/hyperoslo/HYPForms/commit/350d3c1ff57d2c2aca0963f4a147b72067bab7da

Demo/HYPFormsTests/HYPFormsCollectionViewDataSourceTests.m
```diff
@@ -82,6 +82,48 @@ - (void)testIndexInForms
     [self.dataSource processTarget:[HYPFormTarget showFieldTargetWithID:@"last_name"]];
 }
 
+- (void)testEnableAndDisableTargets
+{
+    HYPFormField *targetField = [self.manager fieldWithID:@"display_name" includingHiddenFields:YES];
```
NSElvis: Might be better to have something like this 

```objc
HYPFormField *targetField = [self.manager fieldWithID:@"display_name" includingHiddenFields:YES];
XCTAssertFalse(targetField.disabled);

HYPFormTarget *disableTarget = [HYPFormTarget disableFieldTargetWithID:@"display_name"];
[self.dataSource processTarget:disableTarget];
XCTAssertTrue(targetField.disabled);

HYPFormTarget *enableTarget = [HYPFormTarget enableFieldTargetWithID:@"display_name"];
[self.dataSource processTargets:@[enableTarget]];
XCTAssertFalse(targetField.disabled);
```

zenangst: @NSElvis Agreed, should be fixed now

Demo/HYPFormsTests/HYPFormsCollectionViewDataSourceTests.m
```diff
@@ -82,6 +82,48 @@ - (void)testIndexInForms
...
+
+    [self.dataSource processTargets:@[enableTarget]];
+
+    XCTAssertFalse(targetField.disabled);
+}
+
+- (void)testUpdatingTargetValue
+{
+    HYPFormField *targetField = [self.manager fieldWithID:@"display_name" includingHiddenFields:YES];
+    XCTAssertEqualObjects(targetField.fieldValue, @" ");
```
NSElvis: Wow, I didn't remember this. Shouldn't `targetField.fieldValue` be `nil`?

Demo/HYPFormsTests/HYPFormsCollectionViewDataSourceTests.m
```diff
@@ -82,6 +82,48 @@ - (void)testIndexInForms
...
+
+    HYPFormTarget *updateTarget = [[HYPFormTarget alloc] initWithDictionary:@{
+                                                                              @"id" : @"display_name",
+                                                                              @"type" : @"field",
+                                                                              @"action" : @"update",
+                                                                              @"target_value": @"John Hyperseed"}];
```
NSElvis: Wha is this for? `target_value`

Demo/HYPFormsTests/HYPFormsCollectionViewDataSourceTests.m
```diff
@@ -82,6 +82,48 @@ - (void)testIndexInForms
...
+    [self.dataSource processTarget:updateTarget];
+
+    XCTAssertEqualObjects(targetField.fieldValue, @"John Hyperseed");
+}
+
+- (void)testDefaultValue
+{
+    HYPFormField *usernameField = [self.manager fieldWithID:@"username" includingHiddenFields:YES];
+
+    XCTAssertNotNil(usernameField.fieldValue);
```
NSElvis: Doesn't seem to test anything

Demo/HYPFormsTests/HYPFormsCollectionViewDataSourceTests.m
zenangst: Targets now support setting values to targets.
So if it has a value when you push and update, it will use that target_value on that field.
Try running the Sample and switch username, then you can see it in action.

Demo/HYPFormsTests/HYPFormsCollectionViewDataSourceTests.m
zenangst: Yeah one would think so :)

Demo/HYPFormsTests/HYPFormsCollectionViewDataSourceTests.m
zenangst: Well, it does but the test is funky as it relies on an action in forms.json.
Good that you spotted this because changing the sample would actually break the test.

zenangst: @NSElvis Or maybe that is okay seeing that other tests rely on the forms.json?

zenangst: What it is checking for is that the default value is set to the field if no initial value is supplied
https://github.com/hyperoslo/HYPForms/commit/13e8e926165170319fc93813919209e7e7262766#diff-6038dff43e896530ada273cedc098688R102

Source/HYPFormsCollectionViewDataSource.m
zenangst: @attila-at-hyper actually, it would be like
`_collapsedForms = (_collapsedForms) ? _collapsedForms : [NSMutableArray new];`

zenangst: Or maybe I'm being an idiot here, I'm gonna revert it just to be sure that it does not mess things up big-time :)

zenangst:
Yolo

Source/HYPFormsCollectionViewDataSource.m
NSElvis: I agree, better keep the old one. Nice and clean. No need to do something fancy for something so simple.

Source/HYPFormsCollectionViewDataSource.m
```diff
@@ -451,6 +466,13 @@ - (void)fieldCell:(UICollectionViewCell *)fieldCell updatedWithField:(HYPFormFie
     }
 }
 
+- (void)fieldCell:(UICollectionViewCell *)fieldCell processTargets:(NSArray *)targets
+{
+    dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(HYPFormsDispatchTime * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{
```
NSElvis: I usually use `performSelector:afterDelay` for this, feels more readable

NSElvis:
Nice one! Keep up the great work

PR #141, opened 2015-01-08
Feature/JSON conditionals

zenangst:
This pull request adds conditionals on form targets using `DDExpression`s.
Conditions can be tokenized using values from the values found in the `HYPFormManager`.

Example: `"condition":"$username == 0"`
This update target would only trigger if the username value is equal to "Generated", which has the valueID 0 in a `HYPFieldValue`.


zenangst:
@fespinoza @mathiasbno This is probably interesting for you guys :)

PR #144, opened 2015-01-12
Improve/targets

zenangst:


Source/HYPFormsManager.h
```diff
@@ -42,4 +42,6 @@
 - (NSArray *)enableTargets:(NSArray *)targets;
 - (NSArray *)disableTargets:(NSArray *)targets;
 
+- (BOOL)evaluateCondition:(NSString *)condition;
```
NSElvis: Is this needed?

Source/HYPFormsManager.m
```diff
@@ -450,6 +450,12 @@ - (NSArray *)showTargets:(NSArray *)targets
     for (HYPFormTarget *target in targets) {
         if (![self evaluateCondition:target.condition]) continue;
 
+        __block BOOL fieldIsShown = NO;
+        [self fieldWithID:target.targetID includingHiddenFields:NO completion:^(HYPFormField *field, NSIndexPath *indexPath) {
```
NSElvis: Can you separate parameters in newlines?

Source/HYPFormsManager.m
```diff
@@ -450,6 +450,12 @@ - (NSArray *)showTargets:(NSArray *)targets
...
+            if (field) fieldIsShown = YES;
+        }];
+        if (fieldIsShown) continue;
```
NSElvis: Might be better to wrap the thing in brackets instead of this.

```objc
if (fieldIsHidden) {
// do something
}
```

Source/HYPFormsManager.h
zenangst: Fixed

Source/HYPFormsManager.m
zenangst: Fixed

NSElvis:
What's this code supposed to accomplish?

zenangst:
Avoid showing a field that is already visible. Had a bug where a field got duplicated because of it.
Preventing show on already shown fields fixed it.

Source/HYPFormsManager.m
```diff
@@ -450,51 +450,59 @@ - (NSArray *)showTargets:(NSArray *)targets
...
-            if (field) {
-                HYPForm *form = self.forms[[field.section.form.position integerValue]];
-
-                for (HYPFormSection *section in form.sections) {
-                    if ([section.sectionID isEqualToString:field.section.sectionID]) {
-                        foundSection = YES;
-                        NSInteger fieldIndex = [field indexInSectionUsingForms:self.forms];
-                        [section.fields insertObject:field atIndex:fieldIndex];
+        __block BOOL shouldLookForField = YES;
+        [self fieldWithID:target.targetID includingHiddenFields:NO
```
NSElvis: This should only be call if `if (target.type == HYPFormTargetTypeField) {`

NSElvis:
I've noticed that we are testing hiding a field multiple times, but not showing a field multiple times. Made an issue for that.

PR #145, opened 2015-01-13
Fix/separator width

NSElvis:
Makes the section separator full width


NSElvis:
ios simulator screen shot jan 13 2015 10 33 44 am

PR #146, opened 2015-01-13
Remove info field type

NSElvis:
Fields with typo "info" where used in the JSON but didn't did anything.

### DIFF
--- a/Demo/HYPForms/forms.json
+++ b/Demo/HYPForms/forms.json
@@ -54,7 +54,7 @@
             "id":"display_name",
             "title":"Display name",
             "subtitle":"This is the name that is going to be used across the application.",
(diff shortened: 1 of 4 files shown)


zenangst:
We need to revert this merge, I found where it is being used: https://github.com/hyperoslo/HYPForms/blob/master/Source/Cells/Text/HYPTextFormFieldCell.m#L218

NSElvis:
@zenangst: Would just using `HYPFormFieldTypeText` work?

NSElvis:
@zenangst:   
```objc
BOOL shouldDisplaySubtitle = (self.field.type == HYPFormFieldTypeText && self.field.subtitle);
```

PR #162, opened 2015-01-26
Fix: Default values been nil

NSElvis:
This fixes a bug that caused `default` values been overwritten when initial values contained `NSNull`. See first commit of failing test for more info.


zenangst:
:bug: :skull:

PR #165, opened 2015-01-27
Fix: reloadWithDictionary doesn't run targets

NSElvis:
reloadWithDictionary wasn't running targets, fixed


NSElvis:
Related to #163

Issue #167, opened 2015-01-28
Improve collapsing forms animations


zenangst:
When a form gets collapsed it's contents should flow bellow the header, right now it has goes beyond the header.

NSElvis:
Correct
=====

![correct](https://cloud.githubusercontent.com/assets/1088217/5935717/29cc62b4-a6de-11e4-9642-30f445f5c4df.gif)

Wrong
====

![wrong](https://cloud.githubusercontent.com/assets/1088217/5935719/2eccd406-a6de-11e4-80d5-8c9950e8fcd3.gif)

PR #171, opened 2015-01-28
Fixed size calculation for decoration view

andrewgubanov:
Hi,
This is the fix for the issue " Improve collapsing forms animations #167"
The reason the issue existed was because of the wrong y location which was reset to 0. Y location calculation was also based on the sequence and number of calls for layoutAttributesForDecorationViewOfKind:atIndexPath: which is wrong because this method will not be called for not visible sections even though Y calculation should append the value from it. So I removed dependency of the method calls and this issue is not reproducible anymore.
Please review.

### DIFF
--- a/Source/HYPFormsLayout.m
+++ b/Source/HYPFormsLayout.m
@@ -33,9 +33,6 @@
 
 @interface HYPFormsLayout ()
 
-@property (nonatomic) CGFloat previousHeight;
-@property (nonatomic) CGFloat previousY;
-
 @end
 
 @implementation HYPFormsLayout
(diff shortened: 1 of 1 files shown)


NSElvis:
Hi Andrew, I still see the problem, maybe I haven't explain it well. Here's an screenshot of the current behavior:

ios simulator screen shot jan 28 2c 2015 2c 4 05 37 pm

While the expected behavior looks something like this:

animation

The content of the second form slides up behind the header.

NSElvis:
@andrewgubanov: Hi Andrew! I saw the wrong branch, silly me. Works nicely.

zenangst:
@andrewgubanov :star2:

andrewgubanov:
Thanks everyone for the nice collaboration!

PR #172, opened 2015-01-28
Remove collection view batch update

zenangst:
This caused the app to crash if you try to reload the form with a dictionary and the count of index paths are out for sync. If the current state has more fields than what the reload outcome has.

```Terminating app due to uncaught exception
'NSInternalInconsistencyException', reason: 'attempt to delete item 19
from section 1 which only contains 17 items before the update'```

This fixes that issue and I haven’t seen any side effects from removing
this.


NSElvis:
Would you be able to reproduce the bug in a test? Or would that be to difficult? :hamster:

zenangst:
The problem occurred in the other project but I guess I could give it a shot.

NSElvis:
#yolo

PR #182, opened 2015-01-30
Fix: wrong disabled coloring

NSElvis:
Disabled cells had blue as text color, this was because we were doing it right on popover cells which are Labels, while text cells which were TextFields weren't been updated.


NSElvis:
ios simulator screen shot jan 30 2015 11 39 11 am

NSElvis:
ios simulator screen shot jan 30 2015 11 39 33 am

PR #186, opened 2015-02-02
Add support for dollar signs in formulas

zenangst:


NSElvis:
@zenangst OMG :open_mouth:

NSElvis:
Haha, lol, just saw your implementation, I don't know what to think

NSElvis:
@zenangst I made an issue related to this https://github.com/hyperoslo/NSString-HYPFormula/issues/15

zenangst:
@NSElvis It was done this way to conform to how `$` works in the conditionals. Seeing as how they are optional there, it would make sense to change the entire dictionary structure to make it fit. I might be wrong ;)

NSElvis:
@zenangst I think exactly the same, I just was thinking that `HYPFormula` would be the best place for this kind of implementation.

zenangst:
That is actually true ;)

PR #189, opened 2015-02-03
Provide method to get the number of fields

NSElvis:


NSElvis:
Nice one, we should get more of this, keep the good work :+1:

zenangst:
#yolo

PR #190, opened 2015-02-03
Remove NSNull values from evaluateExpression

NSElvis:
This caused a crash in many places, better not send them, DDExpression
doesn’t play nice with them.


zenangst:
:star2:

PR #191, opened 2015-02-03
Feature/add clear target

zenangst:


Source/HYPFormsCollectionViewDataSource.m
```diff
@@ -511,6 +511,7 @@ - (void)processTarget:(HYPFormTarget *)target
             NSArray *deletedIndexPaths = [self.formsManager hideTargets:@[target]];
             [self deleteItemsAtIndexPaths:deletedIndexPaths];
         } break;
+        case HYPFormTargetActionClear:
```
NSElvis: I guess this needs implementation too, can't remember if we had tests for `processTarget`. =/

```objc
        case HYPFormTargetActionClear: {
            NSArray *updatedIndexPaths = [self.formsManager clearTargets:@[target]];
            [self reloadItemsAtIndexPaths:updatedIndexPaths];
        } break;
```

Source/Models/HYPFormTarget.m
```diff
@@ -274,6 +292,7 @@ + (void)filteredTargets:(NSArray*)targets
             case HYPFormTargetActionHide:
                 if (![hidden containsObject:target]) [hidden addObject:target];
                 break;
+            case HYPFormTargetActionClear:
```
NSElvis: Does `HYPFormTargetActionUpdate` has the same behavior `HYPFormTargetActionClear`?

zenangst: Yeah, it should update the target with a nil value.

Have a look at https://github.com/hyperoslo/HYPForms/blob/feature/add-clear-target/Source/HYPFormsManager.m#L627

I thought it would be smart to share the logic with update instead of duplicating some of the code just to support clearing a field.

Source/HYPFormsCollectionViewDataSource.m
zenangst: But if we don't have a break, doesn't it just hoop down to the next and use update. See last commit for
my way of thinking.

NSElvis:
@zenangst I get it now :) So clear is just a variation of the update target!

PR #193, opened 2015-02-11
Fix/select values not being updated when targets

zenangst:


NSElvis:
OKAY

NSElvis:
@zenangst 

andy-samberg-leave-me-alone

PR #196, opened 2015-02-15
Simplify HYPFormsCollectionViewDataSource initialization

NSElvis:
Before it was a nightmare to initialize and start using HYPForms, now it should be easier

Before you had to hook up delegates, data sources, data sources that had data sources, custom cells where a nightmare ([now is just a block](https://github.com/hyperoslo/HYPForms/blob/simplify-form-delegates/Demo/HYPForms/HYPSampleCollectionViewController.m#L72))
```objc
- (HYPFormsCollectionViewDataSource *)dataSource
{
    if (_dataSource) return _dataSource;

    NSArray *JSON = [NSJSONSerialization JSONObjectWithContentsOfFile:@"forms.json"];

    _dataSource = [[HYPFormsCollectionViewDataSource alloc] initWithJSON:JSON
                                                          collectionView:self.collectionView
                                                                  layout:self.layout
                                                                  values:self.initialValues
                                                                disabled:YES];

    _dataSource.configureCellForIndexPath = ^(HYPFormField *field, UICollectionView *collectionView, NSIndexPath *indexPath) {
        // Return custom cell for field if applies
        // return cell;
    };

    _dataSource.configureFieldUpdatedBlock = ^(id cell, HYPFormField *field) {
        // Field was updated, do something
    };

    return _dataSource;
}
```

**Update**: *Removed the need to create a `formsManager`, this one will be created by the `dataSource` and can be accessed (readonly) doing `dataSource.formsManager`.*

**Update 2**: *Added `disabled` flag.*

### DIFF
--- a/Demo/HYPForms/HYPSampleCollectionViewController.m
+++ b/Demo/HYPForms/HYPSampleCollectionViewController.m
@@ -35,6 +35,13 @@
     self.layout = layout;
     self.initialValues = dictionary;
 
+    [self.collectionView registerClass:[HYPImageFormFieldCell class]
+            forCellWithReuseIdentifier:HYPImageFormFieldCellIdentifier];
+
+    self.collectionView.dataSource = self.dataSource;
+
+    NSLog(@"Initializated with %ld sections", (long)[self.collectionView numberOfSections]);
+
     return self;
 }
 
@@ -59,8 +66,8 @@
     if (_dataSource) return _dataSource;
 
     _dataSource = [[HYPFormsCollectionViewDataSource alloc] initWithCollectionView:self.collectionView
+                                                                            layout:self.layout
                                                                    andFormsManager:self.formsManager];
-    self.layout.dataSource = _dataSource;
 
     _dataSource.configureCellForIndexPath = ^(HYPFormField *field, UICollectionView *collectionView, NSIndexPath *indexPath) {
         BOOL isImageCell = (field.type == HYPFormFieldTypeCustom && [field.typeString isEqual:@"image"]);
@@ -114,12 +121,6 @@
     self.collectionView.contentInset = UIEdgeInsetsMake(20.0f, 0.0f, 0.0f, 0.0f);
 
     self.collectionView.backgroundColor = [UIColor HYPFormsBackground];
-
-    [self.collectionView registerClass:[HYPImageFormFieldCell class]
-            forCellWithReuseIdentifier:HYPImageFormFieldCellIdentifier];
-
-    self.collectionView.delegate = self;
-    self.collectionView.dataSource = self.dataSource;
 }
 
 - (void)viewDidAppear:(BOOL)animated
@@ -160,8 +161,6 @@
     [self setToolbarItems:@[validateButtonItem, flexibleBarButtonItem, updateButtonItem, flexibleBarButtonItem, readOnlyBarButtonItem]];
 
     [self.navigationController setToolbarHidden:NO animated:YES];
-
-    [self.dataSource processTarget:[HYPFormTarget hideFieldTargetWithID:@"image"]];
 }
 
 #pragma mark - UICollectionViewDelegate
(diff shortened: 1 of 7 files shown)


Demo/HYPForms/HYPSampleCollectionViewController.m
```diff
@@ -29,13 +29,19 @@ @implementation HYPSampleCollectionViewController
...
-    layout.dataSource = self;
+    self.layout = layout;
     self.initialValues = dictionary;
 
+    [self.collectionView registerClass:[HYPImageFormFieldCell class]
+            forCellWithReuseIdentifier:HYPImageFormFieldCellIdentifier];
+
+    self.collectionView.dataSource = self.dataSource;
+
+    NSLog(@"Initializated with %ld sections", (long)[self.collectionView numberOfSections]);
```
NSElvis: It's weird, but initializing the controller in a unit test doesn't trigger any of the dataSource methods:

```objc
- (NSInteger)numberOfSectionsInCollectionView:(UICollectionView *)collectionView;
- (NSInteger)collectionView:(UICollectionView *)collectionView numberOfItemsInSection:(NSInteger)section;
```

So the next time these are called is when a change has been made to it's data, so the dataSource no longer contains the correct state.

That's this weird thing is needed.

Maybe would be better to wrap it with a `isTest` call

Demo/HYPForms/HYPSampleCollectionViewController.m
```diff
@@ -29,13 +30,21 @@ @implementation HYPSampleCollectionViewController
...
+    self.layout = layout;
     self.initialValues = dictionary;
 
+    [self.collectionView registerClass:[HYPImageFormFieldCell class]
+            forCellWithReuseIdentifier:HYPImageFormFieldCellIdentifier];
+
+    self.collectionView.dataSource = self.dataSource;
+
+    if ([NSObject isUnitTesting]) {
+        [self.collectionView numberOfSections];
```
NSElvis: It's weird, but initializing the controller in a unit test doesn't trigger any of the dataSource methods:

- (NSInteger)numberOfSectionsInCollectionView:(UICollectionView *)collectionView;
- (NSInteger)collectionView:(UICollectionView *)collectionView numberOfItemsInSection:(NSInteger)section;
So the next time these are called is when a change has been made to it's data, so the dataSource no longer contains the correct state.

That's why this weird thing is needed.

Source/HYPFormsCollectionViewDataSource.m
```diff
@@ -265,11 +271,14 @@ - (CGSize)sizeForItemAtIndexPath:(NSIndexPath *)indexPath
     CGFloat width = 0.0f;
     CGFloat height = 0.0f;
 
-    HYPFormField *field = fields[indexPath.row];
+    HYPFormField *field;
+    if (indexPath.row < fields.count) {
```
zenangst: Safety first! :+1:

Demo/HYPForms/HYPSampleCollectionViewController.m
```diff
@@ -60,9 +69,18 @@ - (HYPFormsCollectionViewDataSource *)dataSource
     if (_dataSource) return _dataSource;
 
     _dataSource = [[HYPFormsCollectionViewDataSource alloc] initWithCollectionView:self.collectionView
+                                                                            layout:self.layout
                                                                    andFormsManager:self.formsManager];
 
-    _dataSource.dataSource = self;
+    _dataSource.configureCellForIndexPath = ^(HYPFormField *field, UICollectionView *collectionView, NSIndexPath *indexPath) {
+        BOOL isImageCell = (field.type == HYPFormFieldTypeCustom && [field.typeString isEqual:@"image"]);
```
zenangst: Could you switch places with this line and line 77 so that the `isImageCell` is coupled to `if (isImageCell)`?

Source/HYPFormsCollectionViewDataSource.m
```diff
@@ -107,7 +112,8 @@ - (NSInteger)collectionView:(UICollectionView *)collectionView numberOfItemsInSe
         return 0;
     }
 
-    return [form numberOfFields:self.formsManager.hiddenSections];
+    NSInteger numberOfItems = [form numberOfFields:self.formsManager.hiddenSections];
```
zenangst: Newline before return?

NSElvis:
@zenangst updated the PR description with something more!

PR #201, opened 2015-02-20
Refactor invalidFormFields

zenangst:


NSElvis:
@zenangst I think so, but it would be more performant at least

NSElvis:
@zenangst I updated the version number too

PR #211, opened 2015-02-25
"Remove" FORMData

NSElvis:
`FORMData` is an implementation detail, the idea is to use `FORMDataSource` as the main entry for the interaction with the data.

In the long run `FORMDataSource` should feel/act very similar to what fetched results controller does, meaning removing/adding fields or sections should _automagically_ update the collection view.

### DIFF
--- a/Source/FORMDataSource.h
+++ b/Source/FORMDataSource.h
@@ -85,4 +85,8 @@ typedef UICollectionViewCell * (^FORMFieldConfigureCellForItemAtIndexPath)(FORMF
 
 - (NSArray *)forms;
 
+- (void)updateValuesWithDictionary:(NSDictionary *)dictionary;
+
+- (NSDictionary *)valuesDictionary;
+
 @end
--- a/Source/FORMDataSource.h
+++ b/Source/FORMDataSource.h
@@ -1,7 +1,7 @@
 @import Foundation;
 @import UIKit;
 
-#import "FORMCollectionViewLayout.h"
+#import "FORMLayout.h"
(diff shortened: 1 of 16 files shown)


Source/FORMDataSource.h
```diff
@@ -56,4 +53,36 @@ typedef UICollectionViewCell * (^FORMFieldConfigureCellForItemAtIndexPath)(FORMF
 - (void)deleteItemsAtIndexPaths:(NSArray *)indexPaths;
 - (void)reloadItemsAtIndexPaths:(NSArray *)indexPaths;
 
+//// FORMData methods
```
zenangst: Do we need this comment?

Source/FORMDataSource.m
```diff
@@ -691,11 +692,87 @@ - (void)formHeaderViewWasPressed:(FORMGroupHeaderView *)headerView
...
 
-#pragma mark - HYPFormsLayoutDataSource
+#pragma mark - FORMLayoutDataSource
 
 - (NSArray *)forms
 {
     return self.formsManager.forms;
 }
 
+//// /////
```
zenangst: :fearful:

Source/FORMDataSource.h
NSElvis: fixed :)

Source/FORMDataSource.m
NSElvis: Fixed :)

zenangst:
:100:

PR #212, opened 2015-02-26
Demo/login

RamonGilabert:


NSElvis:
OMG

RamonGilabert:
It's not done yet... It's just I'm stacked!

Demos/Login-demo/Podfile
```diff
@@ -0,0 +1,5 @@
+source 'https://github.com/CocoaPods/Specs.git'
+
+platform :ios, '7.0'
+
+pod 'Form', git: 'https://github.com/hyperoslo/HYPForms'
```
NSElvis: `pod 'Form', git: 'https://github.com/hyperoslo/Form'`

:)

RamonGilabert: Nice, doing it right now!

Demos/Login-demo/Login-demo/AppDelegate.m
```diff
@@ -0,0 +1,45 @@
...
+
+@implementation AppDelegate
+
+
+- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
+    // Override point for customization after application launch.
+    return YES;
+}
+
+- (void)applicationWillResignActive:(UIApplication *)application {
```
zenangst: Could we remove this boilerplate code?

Demos/Login-demo/Login-demo/JSON.json
```diff
@@ -0,0 +1,48 @@
...
+                    {
+                        "id":"password",
+                        "title":"Password",
+                        "type":"password",
+                        "size":{
+                          "width":60,
+                          "height":1
+                        },
+                        "validations":{
+                          "min-length": 6,
```
zenangst: I believe this should be `min_length`.

RamonGilabert: Oh, you're right!

RamonGilabert:
So basically the first try is, avoiding the problem I have, that is, get the value into a FORMTextField, not a FORMField,  so everytime you update the text from a field, the block gets called and then it checks if everything is correct, if it is, then you tap login and then a UIAlertController pops up. Nice. If not another one pops up. Ideally, what I want to get is, be able to control the buttons from the delegate methods and from the properties in my view controller, if I get this, that's it! :dango:

NSElvis:
*awesome*

RamonGilabert:
So as I was talking with Chris, I think the easiest solution would be to set the alpha to 0.5, that is working in the Demo I'm doing.

A more complete solution, would be, in the ButtonFieldCell class, where you set the alpha to 0 now, set the backgroundColor of the button to gray or another blue instead, and then the enabled property to NO or YES depending on the JSON. It would be 3 lines of code.

NSElvis:
@RamonGilabert Thanks for bringing this up here :)

There's a naming problem, there are two states for a button, hidden and disabled.

Right now disabling a button hides it because we've been treating `disabled` as a `read-mode`, pretty much what iOS Contacts.app does.

In an iOS Contacts.app context if you go into read mode you don't want to see the buttons, and that was our logic. The issue is that we are calling `read-mode` => `disabled`.

Feel free to tweak the behavior of buttons so disable doesn't change the `alpha` to 0, so instead goes to `0.5` (pretty much what you suggested).

In the future we might want to rename the `disabled` property of `FORMDataSource` to something like `editable`, so if `editable` is set to `NO` then buttons will hide.

And when `FORMDataSource` is set to `disabled = YES`  buttons will be alpha 0.5.

RamonGilabert:
Another thing I noticed is that, when you use the method [self.dataSource disable] it works perfect in textFields, but it doesn't in buttons, the buttons are still available to touch. I guess it's because there's no FORMButtonField class, so you're configuring the button as a cell. With that I want to mean, if you want to disable the button, just saw it, you need to disable the cell itself, so the dataSource's disable method won't work.

RamonGilabert:
Right, with this last commit, I say everything is working, the button is disabled at first because of the JSON, and then I make some tweaks to enable it when everything is valid. I had to use the delegate method because the button wasn't listening the touch events always, just the first time. But that's it! :)

NSElvis:
@RamonGilabert would you mind fixing the conflicts?

NSElvis:
screen shot 2015-02-28 at 3 38 36 pm

RamonGilabert:
I've been looking at it, but cannot find where I've caused conflicts, the only thing I touched from the original code is the password support...

RamonGilabert:
Done! :)

NSElvis:
@RamonGilabert Great work! There are some things that could be improved, I'll do it in a different branch for you to review :)

PR #214, opened 2015-02-27
Feature: dynamic sections

NSElvis:
Adds support for templates and dynamic sections.

This feature is compatible with fields even though it's not going to be implemented for now.

The idea is to add a different structure to the `.json` file, while at the same time [supporting the old school way](https://github.com/hyperoslo/Form/commit/daac0e14cf58aaaccff05406f767a7b1a5172b83#diff-dfda6fe51df207387387a8cbe12d7598R124
).

You can add a `templates` dictionary where your fields should have the format ([take a look at the example json](https://github.com/hyperoslo/Form/blob/feature/dynamic-sections/Tests/Tests/JSONs/dynamic.json)):

`"id":"companies[:index].phone_number"`

This will be copied and inserted at the end of the parent section when a `section` with type `dynamic` is pressed.

#### Update: 2 March 2015

The main functionality should be there now. For example [adding and removing dynamic sections](https://github.com/hyperoslo/Form/blob/feature/dynamic-sections/Tests/Tests/HYPFormsCollectionViewDataSourceTests.m#L235) and [generating dynamic sections based on initial values](https://github.com/hyperoslo/Form/blob/feature/dynamic-sections/Tests/Tests/HYPFormsCollectionViewDataSourceTests.m#L284) :)

There's a lot of clean up to be done, but we are getting there.

#### Update: A few minutes after...

I moved the duplicated functionality to FORMData :)


NSElvis:
The main functionality should be there now. For example [adding and removing dynamic sections](https://github.com/hyperoslo/Form/blob/feature/dynamic-sections/Tests/Tests/HYPFormsCollectionViewDataSourceTests.m#L235) and [generating dynamic sections based on initial values](https://github.com/hyperoslo/Form/blob/feature/dynamic-sections/Tests/Tests/HYPFormsCollectionViewDataSourceTests.m#L284) :)

Source/FORMData.m
```diff
@@ -119,13 +137,107 @@ - (void)generateFormsWithJSON:(NSArray *)JSON
...
 
-    [JSON enumerateObjectsUsingBlock:^(NSDictionary *formDict, NSUInteger formIndex, BOOL *stop) {
+    NSArray *groups;
+
+    if ([JSON isKindOfClass:[NSArray class]]) {
+        groups = JSON;
+    } else if ([JSON isKindOfClass:[NSDictionary class]]) {
+        groups = [JSON valueForKey:@"groups"];
+
+        NSDictionary *templates = [JSON andy_valueForKey:@"templates"];
```
NSElvis: There are a lot of strings like this one that could be moved to constants

Source/FORMDataSource.m
```diff
@@ -500,6 +503,70 @@ - (void)fieldCell:(UICollectionViewCell *)fieldCell updatedWithField:(FORMField
         self.configureFieldUpdatedBlock(fieldCell, field);
     }
 
+    NSArray *components = [field.fieldID componentsSeparatedByString:@"."];
```
NSElvis: This is duplicated functionality, I'll do a refactoring before removing the WIP prefix

zenangst:
I think it's easier to do cleanup when we are actually using it. The obvious things should be removed or changed, however... general cleanup I think should go into another pull request. What do you think @NSElvis ?

Source/Models/FORMSection.h
```diff
@@ -4,12 +4,19 @@
 @class FORMGroup;
 @class FORMField;
 
+typedef NS_ENUM(NSInteger, FORMSectionType) {
```
zenangst: :+1:

Source/Models/FORMSection.m
```diff
@@ -33,6 +34,19 @@ - (instancetype)initWithDictionary:(NSDictionary *)dictionary
...
     }];
 
+    if (_type == FORMSectionTypeDynamic) {
+        FORMField *field = [FORMField new];
+        field.position = @(fields.count);
+        field.section = self;
+        field.fieldID = [NSString stringWithFormat:@"%@.add", self.sectionID];
+        field.typeString = @"button";
+        field.type = FORMFieldTypeButton;
+        field.title = [NSString stringWithFormat:@"✚ %@", [self.sectionID uppercaseString]];
```
zenangst: Should this always be `uppercaseString`? Wouldn't it be better to just write upper case in the JSON?

NSElvis: I don't think it works that way =/ 

I was thinking on having a `title` attribute on the section/template, maybe that's the way!

zenangst: Oh, I see!

zenangst: Yeah, having `title` on the `section` would be great!

NSElvis:
Fixes #220

zenangst:
> Qu'ils mangent de la brioche
> - Marie Antoinette

:cake: :cake: :cake: :cake: :cake: :cake:

NSElvis:
@zenangst Thanks!

birthday-kid-is-ready-for-cake-smash

PR #216, opened 2015-02-27
Add AutoCoding

NSElvis:
This enables support for `[instance copy]` I did it in all the classes instead of a "base class" because subclassing hell and also should be easy to overwrite any `copyWithZone`


Source/Models/FORMFieldValue.m
```diff
@@ -50,4 +51,14 @@ - (BOOL)identifierIsEqualTo:(id)identifier
     return NO;
 }
 
+- (id)copyWithZone:(id)zone
+{
+    id copy = [[[self class] alloc] init];
```
zenangst: Newline here?

NSElvis: @zenangst fixed everywhere!

PR #218, opened 2015-03-01
Add new logo

zenangst:


NSElvis:
OMG, looks so pretty! What about adding the name of the thing next to the logo?

NSElvis:
I've found that W: 2500px | H: 650px is the optimal dimension

NSElvis:
@zenangst something like

dqwdqw

zenangst:
I was going for something like this one - https://github.com/guard/guard

NSElvis:
@zenangst I don't like that it breaks the readme and makes it harder to read, specially for us where we have bullet points.

NSElvis:
@zenangst Overall it looks nicer in my opinion: https://github.com/Alamofire/Alamofire

zenangst:
@NSElvis  Yah I agree that might be better for readability! You can go head and add your version to this commit and just merge if you like :)

NSElvis:
@zenangst Should we have the sketch file in the repo?

zenangst:
Why not, I just need to clean the Sketch file so that it's pretty before we do that :+1:

NSElvis:
@zenangst hi!

NSElvis:
@zenangst SOOOO GOOOOOD!

Issue #198, opened 2015-02-18
Custom styling for HYPForms


andrewgubanov:
Please consider implementing possibility to customize style, colors, etc in the HYPForms, because hadrcoded style restricts the usage of this nice library.

zenangst:
Consider it on our todo list :-)

andrewgubanov:
:+1:

NSElvis:
@andrewgubanov :heart:

NSElvis:
We have 

```objc
typedef void (^HYPFieldConfigureCellBlock)(id cell, NSIndexPath *indexPath, HYPFormField *field);
```

It's a configuration block that gets called when the cells get called for usage (reusage), if it's not implemented it only does [this].(https://github.com/hyperoslo/HYPForms/blob/master/Source/HYPFormsCollectionViewDataSource.m#L166)

The idea is that you in this configuration block can change the color of the cell background, cell title label and so on.

You can do the same thing with the `headerConfigurationBlock`.

This might be an overkill since styling should be applied only on initialization.

Other idea could be providing a "THEMEPROVIDER"

Yeah.

andrewgubanov:
@NSElvis I was more thinking about HYPFormBackgroundView which is not possible to style for now.
But anyways, ThemeProvider/ThemeStyler would be the best solution as it could style propper views/cells only when those are instantiated.

NSElvis:
@andrewgubanov maybe considering `UIAppearance` is a good idea too

andrewgubanov:
@NSElvis depends on how much you want to allow user to customize your interface. UIAppearance is good for "coloring" already styled UI. But if you want your users to  be able to actually style the UI, then you need smth like ThemeProvier, etc. It is like if I want to have my backgroundview in a shape of a peer for example. In this case UIAppearance won't help much :)

NSElvis:
I tried UIAppearance with `FORMTextField`, works pretty good.

```objc
[[FORMTextField appearance] setTextColor:[UIColor redColor]];
[[FORMTextField appearance] setBackgroundColor:[UIColor yellowColor]];
```

There are two caveats, the first one is that we need to avoid setting the colors [the way we've been doing it](https://github.com/hyperoslo/Form/blob/b1a542d042a45a9a3056fb8969b5704e51fda1f4/Source/Cells/Views/FORMTextField.m#L98), otherwise appearance doesn't work.

The second one is that I haven't figured out how to style a CALayer using UIAppearance

[...]

PR #222, opened 2015-03-03
[WIP] Custom style

RamonGilabert:
Hey! :)

So which was your idea for the custom styling? I was thinking in the same way you set all the properties, lets say title, subtitle, etc. But for custom styles, backgroundColor, borderRadius, maybe shadow, etc. Right?

### DIFF
--- a/Source/Models/FORMField.h
+++ b/Source/Models/FORMField.h
@@ -43,6 +43,8 @@ typedef NS_ENUM(NSInteger, FORMFieldType) {
 @property (nonatomic) FORMValidationResultType validationType;
 @property (nonatomic) BOOL sectionSeparator;
 
+@property (nonatomic, strong) UIColor *backgroundFieldColor;
+
 - (instancetype)initWithDictionary:(NSDictionary *)dictionary
                           position:(NSInteger)position
                           disabled:(BOOL)disabled


NSElvis:
@RamonGilabert You can get more info here.

https://github.com/hyperoslo/Form/issues/198

The idea is to remove the hardcoded style and move to using `UIAppearance`, and for the layers do something like

```objc
- (void)setCornerRadius:(CGFloat)cornerRadius UI_APPEARANCE_SELECTOR;
```

http://timross.info/blog/2013/03/26/more-fun-with-uiappearance/

RamonGilabert:
@NSElvis, yes, sorry about the `First Commit`, I did it to start the discussion.

Right, so instead of now having this:

```
self.layer.borderWidth = FORMFieldCellBorderWidth;
```

Have this:

```
[[FORMTextField appearance] setBorderWith: x];
```

And, the method to do that will be in the base class. But, for the user to be able to enter information, you want to enter it in the JSON file right? say:

```
"style": {
    "borderRadius": 30
}
```

NSElvis:
@RamonGilabert You can skip the JSON part we don't need that, this can live in the app.

:tropical_drink: :palm_tree:

RamonGilabert:
Ohhhhh! I see, so in the `FORMTextField` we're going to set the original value, then in the app, using one of the blogs we're going to be able to change that value, right? Then everything is the same, just adding those methods.

NSElvis:
@RamonGilabert Do you mind cloning this project and making a PR on it? So I can make changes too? :dancers:

PR #223, opened 2015-03-03
Improve: support multiple dynamic sections in a group

NSElvis:
Adding dynamic sections works perfectly, when a section gets added the position of the sections bellow will be updated `+1`.

But when removing dynamic sections the sections bellow weren't been decreased by one `-1`.

This PR fixes that.


zenangst:
Looks great!

PR #224, opened 2015-03-03
Custom styling

RamonGilabert:
It provides a `FORMDefaultStyle` class that can be invoked from your AppDelegate, it uses UIAppearance to do the custom styling.

The idea is that if you want to provide your own style (in your app) you just copy the default template, rename it and customize it.
 
Fixes https://github.com/hyperoslo/Form/issues/198


NSElvis:
@RamonGilabert I added a `FORMDefaultStyle` class, it might be a good way to provide a default style :)

```objc
- (void)viewDidLoad
{
    [super viewDidLoad];

    [FORMDefaultStyle applyStyle];
}
```

NSElvis:
@RamonGilabert Couldn't help myself onto working on this

RamonGilabert:
@NSElvis sorry, I just got home (2 hours ago) I was out of town, I'll work tomorrow! :)

NSElvis:
@RamonGilabert haha, don't worry :)

NSElvis:
This is the current status

editable
read-mode

RamonGilabert:
You did almost everything already!

RamonGilabert:
So @NSElvis, I found which is the problem, the property is not set to a value when the enable gets called, there's the weak turnaround I did to just call again the enable property with the initial BOOL value, then everything will be correct, my question is, what about if the user wants to set custom values? We should add public methods in the class you created, right?

RamonGilabert:
screen shot 2015-03-05 at 6 56 00 pm

RamonGilabert:
So what I mean is:

`[FORMDefaultStyle setInvalidBackgroundColor:[UIColor redColor]]`

After the `[FORMDefaultStyle applyStyle]`

NSElvis:
@RamonGilabert Ohhh, I don't think we should support that for now, the idea is to get the minimum possible, we can focus on this additional use cases in another pull request :dancers:

NSElvis:
@RamonGilabert Looks like there's one missing thing, the fields should have a gray background when Form is disabled.

RamonGilabert:
Are you thinking in some refactor? Maybe adding some common things like fonts in a class? Or you're good with it? :dancer: @NSElvis!

NSElvis:
@RamonGilabert I thought about it, but having every single thing separated for the default template could be better so you get more control over it.

The idea is that if you want to provide your own style you just copy the default template, rename it and customize it.

NSElvis:
@RamonGilabert Just tested the demo project, I think we nailed it this thing!!! 

tumblr_m6h84maoms1rwcc6bo1_500

NSElvis:
Should be good to review now @hyperoslo/ios !

RamonGilabert:
Nice! :D Yesterday I watched that episode of HIMYM! :D

Source/Cells/Views/FORMFieldHeadingLabel.m
```diff
@@ -1,9 +1,5 @@
 #import "FORMFieldHeadingLabel.h"
 
-#import "UIColor+Hex.h"
-#import "UIFont+FORMStyles.h"
-#import "UIColor+FORMColors.h"
-
 @implementation FORMFieldHeadingLabel
 
 - (instancetype)initWithFrame:(CGRect)frame
```
zenangst: Could we remove this method entirely?

Source/Cells/Views/FORMTextField.m
```diff
@@ -107,14 +115,31 @@ - (void)setEnabled:(BOOL)enabled
...
+        self.layer.borderColor = disabledBackgroundColor.CGColor;
+        self.textColor = disabledTextColor;
+    }
+}
+
+- (void)setValid:(BOOL)valid
+{
+    _valid = valid;
+
+    if (!self.isEnabled) return;
```
zenangst: I think we agree on not having random `return`-statements in methods.

```objc
if (self.isEnabled) {
    if (valid) {
        self.backgroundColor = validBackgroundColor;
        self.layer.borderColor = validBorderColor.CGColor;
     } else {
        self.backgroundColor = [UIColor FORMLightGray];
        self.layer.borderColor = [UIColor FORMFieldDisabledText].CGColor;
        self.textColor = [UIColor grayColor];
        self.backgroundColor = invalidBackgroundColor;
        self.layer.borderColor = invalidBorderColor.CGColor;
     }
}
```

I think that reads better.

Source/Cells/Views/FORMFieldHeadingLabel.m
RamonGilabert: Yes

Source/Others/UIColor+FORMColors.m
```diff
@@ -1,186 +0,0 @@
...
-{
-    return [UIColor FORMRed];
-}
-
-+ (UIColor *)FORMShadowColor
-{
-    return [UIColor colorWithRed: 0.271 green: 0.361 blue: 0.451 alpha: 0.46];
-}
-
-@end
```
zenangst: This is exactly how the red wedding ended.

NSElvis:
@zenangst should be ready!

NSElvis:
@RamonGilabert :dancers:

zenangst:
power-rangers-o

PR #225, opened 2015-03-04
RIP HYPForms, long live FORM

zenangst:


NSElvis:
@zenangst thanks :) 

We might want to do this change for `form` to `group` since what before was a `form` now is a `group`

PR #227, opened 2015-03-04
Fix: demos

NSElvis:
I run `pod update` in both examples and update from `HYPForms` to `FORM` since it's the change it was made a few versions back.


markvaldy:
:+1:

PR #228, opened 2015-03-05
Fix: dynamic sections issue with hidden section

NSElvis:
Hiding section in position `8` should update the position of the sections that are after to `position - 1`.

Showing section with position `8` should update the position of the sections that are after to `position + 1`.


Source/Models/FORMGroup.m
```diff
@@ -118,4 +118,25 @@ - (NSString *)description
...
+            currentSection.position = @([currentSection.position integerValue] - 1);
+        }
+
+        if ([currentSection.sectionID isEqualToString:section.sectionID]) {
+            index = idx;
+            found = YES;
+        }
+    }];
+
+    if (!found) index = [self.sections count];
```
callumboddy: why do you want to remove the last section if section is !found?

NSElvis: @callumboddy you're right, this isn't the expected behavior, that's what I'm trying to accomplish with this PR. We had this as a workaround because we haven't been updating the position of the sections when a section was inserted or removed.

I updated the PR description on what I'm working on :)

PR #232, opened 2015-03-09
Fixed custom font for drop down menus

RamonGilabert:


NSElvis:
Wow, I didn't knew this was needed. It's a bug on UIKit? Or that's how it works?

RamonGilabert:
My guess is that you were calling the setter of set font, because no method was there before I added it!

zenangst:
Okay to merge?

Source/FORMDefaultStyle.m
```diff
@@ -64,7 +64,7 @@ + (void)applyStyle
     [[FORMTextField appearance] setInvalidBackgroundColor:[UIColor colorFromHex:@"FFD7D7"]];
     [[FORMTextField appearance] setInvalidBorderColor:[UIColor colorFromHex:@"EC3031"]];
 
-    [[FORMFieldValueLabel appearance] setFont:[UIFont fontWithName:@"DIN-Regular" size:15.0]];
+    [[FORMFieldValueLabel appearance] setCustomFont:[UIFont fontWithName:@"DIN-Regular" size:15.0]];
     [[FORMFieldValueLabel appearance] setTextColor:[UIColor colorFromHex:@"455C73"]];
```
NSElvis: @RamonGilabert Do we need the same for setTextColor?

zenangst:
Borders on disabled fields are missing.

NSElvis:
@zenangst they look better that way

zenangst:
@NSElvis oh, I didn't know that it was intentional

NSElvis:
@zenangst It wasn't

zenangst:
But it's still a bug that they don't show up? They should still be configurable via `UIAppearance`, right?

NSElvis:
@zenangst it's a bug, I'll make an issue for it

zenangst:
Found it, I'll fix it in another pully :)

PR #249, opened 2015-03-14
Improve basic example

NSElvis:
Show how initial values work


jgorset:
:sparkles: :cake: :sparkles:

PR #260, opened 2015-03-17
Feature/removed dynamic values

zenangst:


Source/FORMDataSource.h
```diff
@@ -65,6 +65,7 @@ typedef UICollectionViewCell * (^FORMFieldConfigureCellForItemAtIndexPath)(FORMF
 - (NSInteger)numberOfFields;
 - (NSArray *)forms;
 - (NSDictionary *)valuesDictionary;
+- (NSDictionary *)removedDynamicValues;
```
NSElvis: Maybe we should make this one `removedValues` too

Source/FORMDataSource.m
```diff
@@ -513,36 +513,60 @@ - (void)fieldCell:(UICollectionViewCell *)fieldCell updatedWithField:(FORMField
...
-                    if ([currentSection.position integerValue] > [section.position integerValue]) {
-                        currentSection.position = @([currentSection.position integerValue] - 1);
+                [self updateSectionPosition:section];
+
+                NSMutableArray *removedKeys = [NSMutableArray new];
+                __block NSDictionary *parsed;
+                [self.valuesDictionary enumerateKeysAndObjectsUsingBlock:^(NSString *key, id obj, BOOL *stop) {
+                    if ([key hasPrefix:section.sectionID]) {
+                        parsed = [key hyp_parseRelationship];
+                        NSString *newKey = [NSString stringWithFormat:@"%@[%@].%@", [parsed objectForKey:@"relationship"], @(self.formsManager.removedValues.count), [parsed objectForKey:@"attribute"]];
```
NSElvis: https://github.com/hyperoslo/NSString-HYPRelationshipParser/issues/7 :)

Source/FORMDataSource.m
```diff
@@ -513,36 +513,60 @@ - (void)fieldCell:(UICollectionViewCell *)fieldCell updatedWithField:(FORMField
...
+                        [self.formsManager.values removeObjectForKey:key];
+                        [removedKeys addObject:newKey];
                     }
-                }
+                }];
+
+                parsed = [[section.sectionID stringByAppendingString:@".name"] hyp_parseRelationship];
```
NSElvis: Appending `.name`?

Source/FORMDataSource.m
```diff
@@ -513,36 +513,60 @@ - (void)fieldCell:(UICollectionViewCell *)fieldCell updatedWithField:(FORMField
...
                 }
+
+                NSDictionary *updatedValueKeys = [self updateValueKeys:[self.valuesDictionary allKeys]];
+                NSDictionary *currentValues = self.valuesDictionary;
+
+                [self removeDynamicKeysForSection:section];
+
+                [currentValues enumerateKeysAndObjectsUsingBlock:^(NSString *key, id obj, BOOL *stop) {
+                    if (updatedValueKeys[key]) {
+                        [self.formsManager.values setObject:obj forKey:updatedValueKeys[key]];
```
NSElvis: I guess 
I think we could use the other notation here

```objc
self.formsManager.values[updatedValueKeys[key]] = obj;
```

Source/FORMDataSource.m
zenangst: :100:

Source/FORMDataSource.m
zenangst: I just used `.name`, no real reference to anything. But `hyp_parseRelationship` wouldn't parse it without the suffix.

Source/FORMDataSource.m
zenangst: Fixed - https://github.com/hyperoslo/Form/commit/6e6d19a003cd28026237d24bb1ed0cfabf2d4307

Source/FORMDataSource.h
zenangst: I thought about that but it's only the dynamic values that gets added to that property to it made sense to make it more explicit. From an API perspective it makes sense but `removedValues` is shorter and reads nice.

NSElvis: Yes please, also in the future adding more functionality to removedValues while been backwards compatible should be easier to do without breaking the API.

zenangst: But for consistency, wouldn't `removedValuesDictionary` be a better name?

NSElvis: I've been thinking that we should go for `values` and `removedValues`, like you don't say

"indexInteger"

zenangst: Fixed :)

zenangst:
:star:

PR #261, opened 2015-03-20
Fix/field positions when using targets

zenangst:


Tests/Tests/JSONs/multiple-show-hide-field-targets.json
```diff
@@ -91,7 +91,7 @@
               },
               {
                 "id":1,
-                "title":"Excelent",
+                "title":"Excellent",
```
NSElvis: :laughing:

PR #264, opened 2015-03-22
Improve/dynamic sections

zenangst:


NSElvis:
I thought the structure of the values would look different so it could be parsed by the nested attributes pod.

https://github.com/hyperoslo/Form/pull/266

zenangst:
I had trouble finding which value to mark as removed, that's why I changed it. Also, the dynamic fields didn't end up in the right location so I refactored that method.

NSElvis:
@zenangst I agree my PR isn't to => develop

My PR is to yours, it needs your changes :)

zenangst:
@NSElvis I just saw that, my bad. *yawn* ... Mondays...

NSElvis:
I thought the structure of the values would look different so it could be parsed by the nested attributes pod.

Look [this](https://github.com/hyperoslo/Form/blob/improve/different-removed-dynamic-values/Tests/Tests/FORMCollectionViewDataSourceTests.m#L360) and [this](https://github.com/hyperoslo/Form/blob/improve/different-removed-dynamic-values/Tests/Tests/FORMCollectionViewDataSourceTests.m#L393) test

Also I found that removing a section isn't updating the other sections properly, [here](https://github.com/hyperoslo/Form/blob/improve/different-removed-dynamic-values/Tests/Tests/FORMCollectionViewDataSourceTests.m#L423)

[Removing section 0 should make section 1 be section 0](https://github.com/hyperoslo/Form/blob/improve/different-removed-dynamic-values/Tests/Tests/FORMCollectionViewDataSourceTests.m#L438)

Source/FORMDataSource.h
```diff
@@ -67,6 +67,7 @@ typedef UICollectionViewCell * (^FORMFieldConfigureCellForItemAtIndexPath)(FORMF
 - (NSArray *)forms;
 - (NSDictionary *)values;
 - (NSDictionary *)removedValues;
+- (void)resetRemovedValues;
```
NSElvis: Does this method need to be public? 

@zenangst

Issue #202, opened 2015-02-20
Check validation types based on `type` instead of `id`, provide default validations


3lvis:

3lvis:
I'll just leave the critical issues open, closing this one.

PR #267, opened 2015-03-23
Improve/input validation

zenangst:
The idea was to decouple the type of field and the input validation.

Phone numbers need to be stored as strings (leading zeros being the issue) but should only contain number. The field is set to `text` but you can override this type by specifying `input` in the validation object.

As an added bonus, the `input` also configures the keyboard.

### DIFF
--- a/Source/Models/FORMField.m
+++ b/Source/Models/FORMField.m
@@ -37,10 +37,6 @@ static NSString * const FORMFormatterSelector = @"formatString:reverse:";
     _type = [self typeFromTypeString:self.typeString];
     _inputType = [dictionary andy_valueForKey:@"input_type"];
 
-    if (_inputType) {
-        _typeString = _inputType;
-    }
-
     NSNumber *width = [dictionary andy_valueForKey:@"size.width"];
     NSNumber *height = [dictionary andy_valueForKey:@"size.height"];
     if (!height || !width) abort();
--- a/Source/Models/FORMField.m
+++ b/Source/Models/FORMField.m
@@ -46,6 +46,10 @@ static NSString * const FORMFormatterSelector = @"formatString:reverse:";
     if (validations && [validations count] > 0) {
         _validation = [[FORMFieldValidation alloc]
                        initWithDictionary:[dictionary andy_valueForKey:@"validations"]];
+
+        if (_validation.input) {
+            _typeString = _validation.input;
+        }
     }
 
     _disabled = [[dictionary andy_valueForKey:@"disabled"] boolValue];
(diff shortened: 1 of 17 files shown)


NSElvis:
I think supporting
```
"type":"phone_numer"
```

Might be a better idea.

https://github.com/hyperoslo/Form/issues/202

zenangst:
What about postal codes and bank account numbers?

NSElvis:
`phone_number` and `email` are generic enough to have `native` support, for everything else you'll use 

```
"validations":{
                "format":"[\\w._%+-]+@[\\w.-]+\\.\\w{2,}",
                "required":true
              }
```

zenangst:
How would this help to specify which keyboard should be used?

Let's say that you have bank account number field:

```JSON
{
  "id":"bank_account_number",
  "type":"text",
  "size":{
    "width":35,
    "height":1
},
  "validations":{
    "input":"0123456789",
    "required":true,
    "min_length":11,
    "max_length":11
  }
}
```

Here you would like it to store it as text (to support leading zeros), so you set the `type` to be `text`.
You then use the `input` key to add allowed characters. What we don't get in this scenario is the right keyboard layout.

If we change the `type` of the field to be `number`, then the keyboard layout is correct but you end up with a different `rawFieldValue` and you lose the ability to store leading zeros.

Am I missing something? :)

NSElvis:
Oh, I see! What about supporting something like:

{
  "id":"bank_account_number",
  "type":"text",
  "input_type":"numeric",
  "size":{
    "width":35,
    "height":1
}

I don't like the keyboard type inside the validations since it's not really a validation.

zenangst:
@NSElvis Yah that makes more sense, I put it inside the validations because it would by proxy only allow the right set of characters but this makes more sense. Let me refactor the PR a bit!

zenangst:
@NSElvis what do you think of this?

Source/Models/FORMField.m
```diff
@@ -35,6 +35,12 @@ - (instancetype)initWithDictionary:(NSDictionary *)dictionary
     _subtitle = [dictionary andy_valueForKey:@"subtitle"];
     _typeString  = [dictionary andy_valueForKey:@"type"];
     _type = [self typeFromTypeString:self.typeString];
+    _inputType = [dictionary andy_valueForKey:@"input_type"];
+
+    if (_inputType) {
+        _typeString = _inputType;
```
NSElvis: This part is odd, can you make it in a way that you don't overwrite the typeString? It's coupled with other logic, meanwhile the inputType is only used for the keyboard

NSElvis:
@RamonGilabert thanks!

PR #268, opened 2015-03-23
Add type safety check to setRawText:

zenangst:


NSElvis:
The bug is related to the dynamic sections setting a weird value, the fix should go there. I'll keep you posted.

PR #277, opened 2015-03-25
Fix/code style

zenangst:


Source/FORMData.m
```diff
@@ -618,7 +626,7 @@ - (NSArray *)showTargets:(NSArray *)targets
         if (target.type == FORMTargetTypeField) {
             [self fieldWithID:target.targetID includingHiddenFields:NO
                    completion:^(FORMField *field, NSIndexPath *indexPath) {
-                       if (field) shouldLookForField = NO;
+                       shouldLookForField = (!field);
```
NSElvis: I'm not sure if this works, for this might be better to `(field == nil)`

zenangst: Good point! I'll fix!

NSElvis: In order to make it no `(field != nil)`

zenangst: @NSElvis fixed in 9090718

NSElvis:
:tada: !

PR #279, opened 2015-03-26
Improve/tooltips

zenangst:
- [x] Rename subtitle view to tooltip view
- [x] Fix bugs related to sticky tool tips when switching between fields
- [x] Add styling for tooltips
- [x] Add tooltip to the JSON
- [x] Animate tool tip


NSElvis:
Just a thought, is `message` better naming than `tooltip`?

Source/Cells/Text/FORMTextFieldCell.m
```diff
@@ -312,10 +327,14 @@ - (void)textFormFieldDidEndEditing:(FORMTextField *)textField
         [self.textField setValid:[self.field validate]];
     }
 
-    [self.subtitleView removeFromSuperview];
+    if (self.showTooltips) {
+    [[NSNotificationCenter defaultCenter] postNotificationName:FORMDismissTooltipNotification
```
NSElvis: This indentation is weird, do you mind fixing it?

zenangst:
I dunno, I kinda like `tooltip´.

Source/Cells/Text/FORMTextFieldCell.m
zenangst: Fixed 4e16e2a

NSElvis:
Makes sense that the UI component is tooltip, but in the JSON maybe it should be `message`.

What do you think @fespinoza?

zenangst:
Yah that I can agree on :+1:

zenangst:
The tooltip is an implementation detail but the JSON should be agnostic to that.
My vote is on `message` for the JSON

zenangst:
@NSElvis fixed!

NSElvis:
Great work!!! 

minion

PR #286, opened 2015-03-27
Fixed #285. If it's enabled initially it's going to show tooltips

### LINKED ISSUE #285: Tooltip not showing up when initializating FORMDataSource as enabled

RamonGilabert:


NSElvis:
Just tried this, it doesn't solve the problem. I'll check what's going on.

NSElvis:
@zenangst I've been collaborating with @RamonGilabert on this one.

In my opinion it doesn't make sense that the tooltip gets shown in disabled fields at all, I can't see the scenario of someone trying to see the message behind a disabled field.

Also this fixes the issue with the tooltip not been shown when FORMDataSource gets initialized as enabled.

zenangst:
We have this need in our current application.

NSElvis:
Why?

zenangst:
Simple, we have a need to display a message saying that you cannot edit this field yourself but if the information is incorrect you can contact support.

zenangst:
😎

NSElvis:
It doesn't make sense right now, since it's just disabled but when I thought we could add an "info" indicator meaning this field contains additional information then it makes a *lot* of sense.

zenangst:
Agreed, do the same thing as we did with separating tooltip and subtitle?

PR #287, opened 2015-03-28
func to invalidate fields

matteocrippa:
dunno if I missed something but I was interested in invalidating all the form resetting all the fields to empty status, so I managed to add this quick func


NSElvis:
Hi @matteocrippa! 

By any chance have you tried `- (void)resetForms;`? 

https://github.com/matteocrippa/Form/blob/master/Source/FORMDataSource.h#L38

:smiley:

matteocrippa:
@NSElvis yep, but it was totally removing the fields from the viewcontroller, my need was only eg to reset the UILabel to nil

NSElvis:
Hi @matteocrippa, you're totally right, the `resetForms` is not acting as it should. I'll move your implementation to that method so you can use it.

Thanks!

3lvis:
Haha, wth.

matteocrippa:
ops

PR #291, opened 2015-03-30
Refactor: rename forms to groups

NSElvis:
A Form is not a group, because FORMForm sounded weird.

This is the final renaming that started when we went 1.0


zenangst:
axb6mlz_460sa_v1

Issue #219, opened 2015-03-02
Section where fields don't add up to 100 have weird behaviours


3lvis:
Example:

First name field => 50 : 1
Last name field => 25 : 1
Image field => 100 : 1.7

zenangst:
@NSElvis mind adding a screenshot?

NSElvis:
@zenangst What should be the behaviour here? Crash and burn?

zenangst:
@NSElvis if the JSON is served up from a backend it could be risky to force a crash, what do you think?

NSElvis:
A crash when parsing the field that is adding up more than 100 and saying "Why?" could be a good solution.

Backends should calculate correct widths, we can't really handle all the scenarios.

What do you think @fespinoza?

zenangst:
Oh sorry.

richardoti:
@zenangst should we close?

3lvis:
Haven't hard of anyone finding this critical. Closing and let's reopen if this becomes critical.

PR #301, opened 2015-03-31
Improve layout rendering

zenangst:
By taking bounds into account, this should catch fields that try to render themselves off screen by resetting their x position.

- [x] Fix out of bound fields
screen shot 2015-03-31 at 13 06 21

- [x] Fix height offset in decoration view

**Before**
before

**After**
after

- [x] Clean up


zenangst:
This PR should fix https://github.com/hyperoslo/Form/issues/219

NSElvis:
Awesome!!!

We should introduce unit testing for our layouts soon :)

zenangst:
@NSElvis very true! :)

PR #310, opened 2015-04-06
Update swift demo

kostiakoval:
1 - don't use self.
2 - Lazy properties
3 - Double numbers format
4 - Split code in extensions

5 - Crazy idea to keep class variable declaration clean by making Lazy struct to init lazy property
this idea could sound little crazy :)
it is a mix of Lazy loading mechanism and keeping Class declaration clean

https://medium.com/swift-programming/how-to-do-proper-lazy-loading-in-swift-b8bc57dbc7b9
and this 
https://medium.com/swift-programming/keep-your-swift-models-clean-ea7eb314d49d

### DIFF
--- a/Demos/Basic-Swift/Basic-Swift/SampleCollectionViewController.swift
+++ b/Demos/Basic-Swift/Basic-Swift/SampleCollectionViewController.swift
@@ -2,18 +2,25 @@ import UIKit
 
 class SampleCollectionViewController: UICollectionViewController {
 
-  lazy var dataSource: FORMDataSource = Lazy.dataSource(self)
   var layout: FORMLayout
 
   var initialValues :Dictionary<NSObject, AnyObject>?
   var JSON: AnyObject?
+
+  lazy var dataSource: FORMDataSource = {[unowned self] in
+    FORMDataSource(JSON: self.JSON,
+      collectionView: self.collectionView,
+      layout: self.layout,
+      values: self.initialValues,
+      disabled: true)
+    }()
 
   init(initialValues: Dictionary<NSObject, AnyObject>, JSON: AnyObject?) {
     self.initialValues = initialValues
     self.JSON = JSON
     layout = FORMLayout()
 
-    super.init(collectionViewLayout: self.layout)
+    super.init(collectionViewLayout: layout)
   }
 
   required init(coder aDecoder: NSCoder) {
@@ -100,20 +107,6 @@ class SampleCollectionViewController: UICollectionViewController {
       "username": 1
       ])
     collectionView?.reloadData()
-  }
-}
-
-extension SampleCollectionViewController  {
-
-  struct Lazy  {
-
-    static func dataSource(aSelf: SampleCollectionViewController) -> FORMDataSource {
-      return FORMDataSource(JSON: aSelf.JSON,
-        collectionView: aSelf.collectionView,
-        layout: aSelf.layout,
-        values: aSelf.initialValues,
-        disabled: true)
-    }
   }
 }
 
--- a/Demos/Basic-Swift/Basic-Swift/SampleCollectionViewController.swift
+++ b/Demos/Basic-Swift/Basic-Swift/SampleCollectionViewController.swift
@@ -2,7 +2,7 @@ import UIKit
 
 class SampleCollectionViewController: UICollectionViewController {
 
-  var formDataSource: FORMDataSource?
+  lazy var dataSource: FORMDataSource = Lazy.dataSource(self)
   var layout: FORMLayout
 
   var initialValues :Dictionary<NSObject, AnyObject>?
@@ -11,29 +11,15 @@ class SampleCollectionViewController: UICollectionViewController {
   init(initialValues: Dictionary<NSObject, AnyObject>, JSON: AnyObject?) {
     self.initialValues = initialValues
     self.JSON = JSON
-    self.layout = FORMLayout()
+    layout = FORMLayout()
 
     super.init(collectionViewLayout: self.layout)
   }
 
   required init(coder aDecoder: NSCoder) {
-    self.layout = FORMLayout()
+    layout = FORMLayout()
 
     super.init(coder: aDecoder)
-  }
-
-  // MARK: Getters
-
-  func dataSource() -> FORMDataSource {
-    if self.formDataSource == nil {
-      self.formDataSource = FORMDataSource(JSON: self.JSON,
-        collectionView: self.collectionView,
-        layout: self.layout,
-        values: self.initialValues,
-        disabled: true)
-    }
-
-    return self.formDataSource!
   }
 
   // MARK: View life cycle
@@ -41,9 +27,9 @@ class SampleCollectionViewController: UICollectionViewController {
   override func viewDidLoad() {
     super.viewDidLoad()
 
-    self.edgesForExtendedLayout = .None
-    self.collectionView?.backgroundColor = UIColor(fromHex: "DAE2EA")
-    self.collectionView?.dataSource = self.dataSource()
+    edgesForExtendedLayout = .None
+    collectionView?.backgroundColor = UIColor(fromHex: "DAE2EA")
+    collectionView?.dataSource = dataSource
   }
 
   override func viewDidAppear(animated: Bool) {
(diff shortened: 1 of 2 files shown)


kostiakoval:
What do you think @zenangst

Demos/Basic-Swift/Basic-Swift/SampleCollectionViewController.swift
```diff
@@ -52,64 +38,60 @@ class SampleCollectionViewController: UICollectionViewController {
...
+    let flexibleBarButtonItem = UIBarButtonItem(barButtonSystemItem: .FlexibleSpace)
 
-    let readOnlyView = UIView(frame: CGRectMake(0.0, 0.0, 150.0, 40.0))
-    let readOnlyLabel = UILabel(frame: CGRectMake(0.0, 0.0, 90.0, 40.0))
+    let readOnlyView = UIView(frame: CGRectMake(0, 0, 150, 40))
+    let readOnlyLabel = UILabel(frame: CGRectMake(0, 0, 90, 40))
     readOnlyLabel.text = "Read-Only"
     readOnlyLabel.textColor = UIColor(fromHex: "5182AF")
-    readOnlyLabel.font = UIFont.boldSystemFontOfSize(17.0)
+    readOnlyLabel.font = UIFont.boldSystemFontOfSize(17)
```
zenangst: Why did you drop the `.0` when `boldSystemFontOfSize` accepts floats? :)

zenangst:
1. I guess using `self` is a habit that comes from Objective-C. I'm torn what I think about dropping it as it clearly state that it's an instance property and not something that is confined to the current scope.

2. :heart: 

3. What's the reasoning behind this? I tend to follow what is stated in the docs.

4. Could you end up with some odd behaviour if you fill your application with extensions? I like the idea.

5. I like it, however.. I wanted the Demo to be vanilla without sprinkling too much magic on it so that the implementation doesn't steal focus from how you actually use Form in your project. Do you agree?
I might be crazy :dizzy_face:

3lvis:
I'm torn on the `self` part too, I don't see how dropping it would make the code more clear. 

Less code isn't always better code.

Demos/Basic-Swift/Basic-Swift/SampleCollectionViewController.swift
kostiakoval: it's still Double, but because of swift type inference, you don't need to specify ,decimal notation ".0"
example:
All are Doubles, same
```swift 
let a: Double = 17
let b = 17.0
var c: Double
c = 17
```

3lvis: I like to think that the `.0` never was intended for the compiler, but for other readers.

kostiakoval:
1 - when you drop using self, Xcode warn you sometimes when you must use self (example in closure, and this help avoiding retain cycles, strong self referencing)

2 - makes code cleaner, less word and letters, easier to read.
self.sendMessage vs sendMessage

kostiakoval:
Agree on point 4, it's to advanced. I'll change it t default lazy property style

3lvis:
But makes it harder to differentiate between global variables and local ones.

zenangst:
@kostiakoval 
2. I agree that it reads better, but in a large method, you quickly lose the ability to see if it's something that is declared in the current scope or not. It makes it harder to get a quick overview when viewing it without 100% context.

kostiakoval:
you get Xcode highlight them differently.

I've seen many recommend doing it
https://github.com/raywenderlich/swift-style-guide#use-of-self
https://github.com/raywenderlich/swift-style-guide#use-of-self

Demos/Basic-Swift/Basic-Swift/SampleCollectionViewController.swift
zenangst: I think it's much easier to be consistent if you use the same type as stated in the docs.
Unless we agree to do something else in our guidelines.

kostiakoval:
It's totally very different from objective-c and It will fill weird and painful at first.

zenangst:
Oh right, Xcode does highlight them differently... well duh! Now I feel like a dumb-ass for not noticing that earlier :)

3lvis:
But you also read code in other places, you shouldn't be bound to jump to Xcode for reading code.

zenangst:
@3lvis very good point!

kostiakoval:
I agree with @3lvis point. 
But still probably there should be some point it that we don't see now. It's not coincidence that it exist and people use it. It just feel good :)

kostiakoval:
"The reason I like not using self is that when closure capture forces you to use self, it then stands out (which is good, because referencing self in a closure retains it, unlike other contexts, so you want it to look "different") –  Catfish_Man J"

kostiakoval:
@zenangst 
> Could you end up with some odd behaviour if you fill your application with extensions? I like the idea.

Nope you shown't.

Demos/Basic-Swift/Basic-Swift/SampleCollectionViewController.swift
kostiakoval: it's ok for me to use 17.0 , and .0 for all float and double numbers. 

But i feel in swift it's more about cleaner and removing distraction 
you say what you want not ho: 
I want font of size 17 
I want a font of size with Float number 17.0.
P.S the same is valid for using `self`

kostiakoval: from raywenderlich styleguide - 
`let bounds = CGRect(x: 40, y: 20, width: 120, height: 80)`

3lvis: Is that integers or floats?

kostiakoval: float, because of type inference, you don't need to specify it, compiler understands it
CGRect init(x: CGFloat, y: CGFloat, width: CGFloat, height: CGFloat)

3lvis: Is not all about the compiler, code should be readable in all contexts, even in plain file formats.

zenangst: I'm with @3lvis on this one. When you read someone else's code, you usually do it on GitHub and not in Xcode. So it needs to be explicit enough in that context for it to make sense.

kostiakoval: see my comment above .
It's not for compiler but for people who are reading code.
I read this code - 
CGRect(x: 40, y: 20, width: 120, height: 80) - make a rect with numbers
CGRect(x: 40.0, y: 20.0, width: 120.0, height: 80.0) - make a rect with float numbers. It adds extra information to process that I don't need. 

Express the intention and remove all unnecessary information.
CGRect(x: 40, y: 20, width: 120, height: 80) - this one read Perfect in txt file, without any code higgling.

kostiakoval: @3lvis @zenangst 
> Is that integers or floats?
You shouldn't think about it when reviewing code, because Swift don't allow you to use wrong type. :)

kostiakoval: How about that one ?
let name =  getName()
not specifying type explicitly

3lvis: I agree with you, specifying that it's a float or an integer doesn't add any value for this use case but there are times where you would have to use the `.` and inconsistency will be introduced.

Example A:
```swift
let space = 2.2
let width = space/1.0 // Double
```

Example B:
```swift
let space = 2
let width = space/1.0 // Error, you need to put `let space = 2.0`
```

kostiakoval: yes sure. 
Example B MUST be
```swift
let space = 2.0 
```
But you could divide it by `1` 
```swift
let width = space / 1
```

3lvis: Diving it by 1 would make it integer

kostiakoval: nope :)

kostiakoval: 1 - is Double in that example, because of type inference 
screen shot 2015-04-06 at 22 43 06

3lvis:
```swift
let space = 2
let width = space/1
```

kostiakoval: in that case everything is `Int`. and it's very clear for me.

3lvis: @kostiakoval I think you're missing the point. We are trying to achieve consistency and have a code style that has the least amount if "except when".

kostiakoval: `1` has the same meaning as `1.0` for humans 
`2 / 1` == `2 / 1.0`, my suggestion, skip `.0` when it's possible
``` swift
let part = 2.0 //here not possible to skip .0 part 
let myPart = part / 2
```

kostiakoval: Rule 
- use .0 when declaring a Doulble or float type and assigning a value.  
- don't use is there is type inference available  
consistency :)

kostiakoval: This is my thought, you decide :) :tada: :dancers: :cake: ))

3lvis: This is a simpler rule and much easier to follow:
- Use .0 when declaring a double or float type

kostiakoval: +
- Use `.x` when you specify float part, not != 0. Example: `UIFont(11.4)`

Demos/Basic-Swift/Basic-Swift/SampleCollectionViewController.swift
```diff
@@ -52,64 +45,60 @@ class SampleCollectionViewController: UICollectionViewController {
...
       style: .Done,
       target: self,
-      action: NSSelectorFromString("validateButtonAction"))
+      action: "validateButtonAction")
 
     let updateButtonItem = UIBarButtonItem(title: "Update",
       style: .Done,
       target: self,
-      action: NSSelectorFromString("updateButtonAction"))
+      action: "updateButtonAction")
```
3lvis: Is removing `NSSelectorFromString` better?

Demos/Basic-Swift/Basic-Swift/SampleCollectionViewController.swift
```diff
@@ -117,14 +106,20 @@ class SampleCollectionViewController: UICollectionViewController {
...
   func collectionView(collectionView: UICollectionView!, layout collectionViewLayout: UICollectionViewLayout!, sizeForItemAtIndexPath indexPath: NSIndexPath!) -> CGSize {
-    return self.dataSource().sizeForFieldAtIndexPath(indexPath)
+    return dataSource.sizeForFieldAtIndexPath(indexPath)
   }
-
 }
 
+extension UIBarButtonItem {
+
+  convenience init(barButtonSystemItem systemItem: UIBarButtonSystemItem) {
```
3lvis: :O

Demos/Basic-Swift/Basic-Swift/SampleCollectionViewController.swift
kostiakoval: yes, it says do to this way in Apple Doc.

Demos/Basic-Swift/Basic-Swift/SampleCollectionViewController.swift
kostiakoval: :) clean code, don't pass nil, when not necessary, use fault parameters

Demos/Basic-Swift/Basic-Swift/SampleCollectionViewController.swift
kostiakoval: https://developer.apple.com/library/ios/documentation/Swift/Conceptual/BuildingCocoaApps/InteractingWithObjective-CAPIs.html#//apple_ref/doc/uid/TP40014216-CH4-XID_39

3lvis: Amazing!

3lvis:
I'll just kill this for now, we'll revisit the sample project in the future.

PR #312, opened 2015-04-07
Improve/api

zenangst:
- [x] Expose collapseFieldsInSection:collectionView:
- [x] Add sectionIsCollapsed
- [x] Add support for custom headers
- [x] Add tests


3lvis:
Is this WIP?

zenangst:
@3lvis yah, sorry. Fixed now

Source/FORMDataSource.m
```diff
@@ -752,6 +756,15 @@ - (void)insertedIndexPathsAndSectionIndexForSection:(FORMSection *)section
     }
 }
 
+- (BOOL)sectionIsCollapsed:(NSInteger)section
```
kostiakoval: this should work `return [self.collapsedGroups containsObject:@(section)]`

Source/FORMDataSource.h
```diff
@@ -207,6 +207,12 @@ includingHiddenFields:(BOOL)includingHiddenFields
                                 NSArray *indexPaths))completion;
 
 /*!
+ * @return @c YES if the Section is collapsed.
+ * @param sectionID The section index for the section.
+ */
+- (BOOL)sectionIsCollapsed:(NSInteger)section;
```
3lvis: I don't think this is the right naming.

This should be

```objc
- (BOOL)groupIsCollapsed:(NSInteger)groupID;
```

Source/FORMDataSource.h
```diff
@@ -217,4 +223,11 @@ includingHiddenFields:(BOOL)includingHiddenFields
  */
 - (void)reloadFieldsAtIndexPaths:(NSArray *)indexPaths;
 
+/*!
+ * @param section A section index for the current section
+ * @param collectionView The used collectionView, usually from a @c UICollectionViewController.
+ */
+- (void)collapseFieldsInSection:(NSInteger)section
```
3lvis: I don't think this is the right naming.

This should be

```objc
- (void)collapseGroup:(NSInteger)groupID; (we don't need the collectionView)
```

3lvis:
@zenangst Would you mind adding a few tests for this?

Source/FORMDataSource.h
zenangst: I like it, but it's not really the group ID, it's the group index.

3lvis: Might be better if we make it use the group ID instead of the index.

zenangst:
@3lvis of course! :)

Source/FORMDataSource.h
3lvis: @zenangst Read comments above

3lvis: @zenangst Doesn't seem that you need this method to be public.

Source/FORMDataSource.h
```diff
@@ -207,6 +218,12 @@ includingHiddenFields:(BOOL)includingHiddenFields
                                 NSArray *indexPaths))completion;
 
 /*!
+ * @return @c YES if the Group is collapsed.
+ * @param group The group index for the group.
+ */
+- (BOOL)groupIsCollapsed:(NSInteger)group;
```
3lvis: @zenangst Doesn't seem that you need this method to be public.

3lvis:
Fantastic job as always

PR #322, opened 2015-04-13
Improve/when used as framework

zenangst:
Use `[NSBundle mainBundle]` instead of `bundleForClass:`


kostiakoval:
Are you sure main bundle will work in framework? I don't think so.

zenangst:
@kostiakoval works just fine in my current project where I'm using this branch :)

PR #330, opened 2015-04-16
Fix popover validation failing

### LINKED ISSUE #329: Validation Issue
https://github.com/hyperoslo/Form/issues/327

Can someone solve it?

Regards!

3lvis:
self.fieldValueLabel’s `setValid:` received a BOOL,

And `[self.field validate]` was sending a `FORMValidationResultType`

Fixes #329

### DIFF
--- a/Source/Cells/Popover/FORMPopoverFieldCell.m
+++ b/Source/Cells/Popover/FORMPopoverFieldCell.m
@@ -78,7 +78,8 @@ static const CGFloat FORMIconButtonHeight = 38.0f;
 
 - (void)validate
 {
-    [self.fieldValueLabel setValid:[self.field validate]];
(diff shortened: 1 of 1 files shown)


Mutix:
Good catch, just came across this myself.

Is the rest of the project in a state where an update to the pod can be pushed?

3lvis:
Yes, this bug was introduced on a previous PR. The latest release is stable and not affected by this.

So running `pod update Form` should be fine, unless you need support for adding minimum of maximum value validations.

Mutix:
Thnaks 3lvis,

This bug is actually present in the current 3.2.1 release as well.

3lvis:
Just made a new release :)

Mutix:
Awesome, cheers :)

PR #337, opened 2015-04-16
Keep null values in the evaluator

zenangst:
I tried adding a rule to a dynamic field but those do not show up in
the dictionary because they are added without an initial value (NSNull).

### DIFF
--- a/Source/FORMData.m
+++ b/Source/FORMData.m
@@ -949,23 +949,17 @@ includingHiddenFields:(BOOL)includingHiddenFields
     }
 }
 
-- (BOOL)evaluateCondition:(NSString *)condition {
+- (BOOL)evaluateCondition:(NSString *)condition
+{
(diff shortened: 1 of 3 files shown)


zenangst:
Also, you might want to add your own evaluator that something shouldn't be null and so on.

3lvis:
There was a reason why this was added. Can't remember at the moment. Have you tried this in your app without any issues? Do you prefer leaving adding tests for this to a new task, or you want to include some?

zenangst:
So far I haven't seen any issue with it, might have been so that you could look for keys instead of values when doing `empty`. Gonna investigate it further tomorrow.

zenangst:
@3lvis I think this is the reason why they were not included, fixed in https://github.com/hyperoslo/Form/commit/42a24bf980e3648b96d26447733d1d4f8e2d894a

PR #339, opened 2015-04-16
Included the field icons in the pod resource bundle 

### LINKED ISSUE #334: Include the field icons in the pod resource bundle
Currently, the icons for the down arrow, the calendar... and soon the clock ;) are included as resources of the demo app, rather than being bundled with the pod.

When using the pod 'out of the box' the images are missing and need to be manually added to the client project.

Would be nice for the images to be directly bundled with the pod :+1:

Mutix:
Fixes #334 .

This change has removed the field icon images (`ic_calendar@2x`, `ic_clock@2x`, `ic_mini_arrow_down@2x` and `ic_mini_clear@2x`) from the demo projects and added them within a resource bundle packaged with the pod.

All future images used by classes within the Form component should be added to the /Assets folder at the root of the library directory. They will automatically be bundled up by cocoapods when running pod install and be included in a resource bundled defined in the podspec (Form.bundle).

To use them inside the classes, the image names need to be prefixed with the bundle.

For example, `[UIImage imageNamed:@"ic_calendar"]` becomes

`[UIImage imageNamed:@"Form.bundle/ic_calendar"]`

More info on cocoapods resource bundles [here](https://guides.cocoapods.org/syntax/podspec.html#resource_bundles).

Changelog:

- deleting field icon images from swift demo project
- deleting field icon images from the ObjC demo
- added field icon images to new Assets folder at root of the library
- removed references to icons in the objc and swift demo projects
- added the resource bundle declaration in the podspec
- changed the initialisation of the field icon images to use the Form.bundle in FORMDateFieldCell, FORMSelectFieldCell and FORMTextField


3lvis:
Wow!

3lvis:
Great job @Mutix! I'm totally doing this is all my pods now.

happy-gof

PR #343, opened 2015-04-17
Refactor to use isEqual:

zenangst:


Source/Others/DDMathEvaluator+FORM.m
```diff
@@ -38,7 +38,9 @@ + (NSDictionary *)hyp_directoryFunctions {
 
             otherValue = (variables[expression.variable]) ?: expression.variable;
 
-            if (![baseValue isEqualToString:otherValue]) {
+            BOOL baseValueIsEqual = ([baseValue isKindOfClass:[NSString class]] &&
+                                     ![baseValue isEqualToString:otherValue]);
```
3lvis: What about numbers? Should we add another `if` in case it's a number?

zenangst: You don't need `equals` to support numbers. You could do `eq($age,15)` or just `$age == 15` (I think).

zenangst: For reference : https://github.com/davedelong/DDMathParser/wiki/Built-in-Functions

zenangst: ping @3lvis @vadymmarkov

3lvis:
```objc
BOOL baseValueIsEqual = (([baseValue isKindOfClass:[NSString class]] &&
                         ![baseValue isEqualToString:otherValue]) || 
                         ([baseValue isKindOfClass:[NSNumber class]] &&
                         ![baseValue isEqualToNumber:otherValue]));
```

zenangst: My question is, should it be supported?

3lvis: I was wondering too :)

3lvis:
I guess I will just merge.

zenangst:
@3lvis I would like to add some tests with null values

zenangst:
@3lvis I think it should be good now

PR #346, opened 2015-04-17
Improved Login demo app

### LINKED ISSUE #345: Add support for secure entry text fields (eg. passwords)
Would be awesome to have a new 'secure' field type which would set  `secureTextEntry` to true on the UITextField.

The login demo app can then use this new type for the password field.

Will start a branch in my fork :)

Mutix:
Tweaked the Login demo app to use the password input_type (the password is now secure text).

Also fixed a bug where if you entered valid data in both fields, the login button would be enabled, however if you then cleared one of the fields (making the form invalid), the login button would not be disabled.

Can now close #345 .

Changes:

- added password input_type to the password field in the JSON
- refactored the HYPDemoLoginCollectionViewController slightly
- fixed an issue where the login button would not be disabled after a the form becomes invalid after it  was valid at least once

### DIFF
--- a/Demos/Login-demo/Login-demo/HYPDemoLoginCollectionViewController.m
+++ b/Demos/Login-demo/Login-demo/HYPDemoLoginCollectionViewController.m
@@ -84,32 +84,42 @@
 
 - (void)fieldCell:(UICollectionViewCell *)fieldCell updatedWithField:(FORMField *)field
 {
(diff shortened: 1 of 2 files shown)


3lvis:
Looks awesome. :clap:

3lvis:
Thanks @Mutix!

PR #353, opened 2015-04-17
FORMLayout using the UICollectionView's bounds to calculate sizes instead of the main screen bounds, enabling using forms in views which are not full screen

### LINKED ISSUE #350: Add ability to set custom bounds for forms which are not full-screen
Currently, the bounds used by the form layout are the main screen's bounds.

If we attempt to present a form in a modal form sheet presentation style on the iPad, the form layout is wrong and the width for groups, sections and fields end up being too large for the reduced view size.

For example:

ios simulator screen shot 17 apr 2015 14 08 21

 
A potential solution could be exposing a `formBounds` property in the FORMLayout class, and using this instead of the main screen bounds for calculating sizes.

A new `initWithFormBounds:` initialiser could also be added.

Any thoughts on how best to implement this?

Mutix:
Fixes #350 

Change-log below. As far as I can see, rotation handling and tests are unaffected.

- changed the FORMLayout to use the collection view's bounds instead of the main screen's bounds to calculate sizes
- changed the FORMDataSource to use the collection view's bounds instead of the screen bounds to calculate field sizes
- now displaying a navigation bar in the obj-c sample app
- added a bar button in the navigation bar to launch a form sheet style modal
- added a new HYPSampleModalFormController to display a sample form in a form sheet style modal
- added a new modal_form.json file to use in the modal form example (same as the original but with different widths for the employment date and bonus fields)
- removed dependency on UIScreen-HYPLiveBounds since it is no longer used
- fixed header label frame calculation in FORMGroupHeaderView which was using UIScreen's bounds (without liveBounds), now using the header's own bounds
- added nice form logo to the login example


3lvis:
Hi @Mutix,

Fantastic work on this pull request! I'm so happy that we finally got rid of `hyp_liveBounds`. 

happy-excited-gif

3lvis:
@Mutix Mentioned you on our last release, thanks for all your work.

https://github.com/hyperoslo/Form/releases/tag/3.3.1

Mutix:
Thanks guys, much appreciated - happy to help when I can :+1:

PR #377, opened 2015-04-22
Fix: remove hidden values and add shown values

### LINKED ISSUE #359: Sub Questions Update/Save Issue
Hello everyone!
I created my form with a save button,
I used this code to log the results i want to save,
```Objective-C
- (void) saveSurvey
{
    NSDictionary *valuesDictionary = [self.dataSource values];
    NSLog(@"%@",valuesDictionary);
}
```
In this Gif i show an issue that if i have a field of type selection that has 3 options, the first one shows a subquestion, the second one shows another subquestion, the last one show another subquestion also.
I hit the save every time when change the selected value, as you can see in the log, the old answer of the old subquestion is not removed, so my survey result must contains 9 answers which is in the first log,
then it became 10, then 11 because when changing the selection from dropdown menu, the subquestion answer is not removed. Hope my explanation is clear.
Those are the ids of each field by name.
1	Activity Type
2	Date of Activity
3	Topic of the Activity
4	Speaker
5	Targeted Products
7	Additional Comments and observations
100	Rate Experience
101	Why?
102	Ok, Why?
103	Select One
104	Sex
105	Why?

if rate experience was great, 'select one' is one. when changed rate experience to average and put 'OK, Why' to Dunno and hit the save, 'Select One' question must be set to null.
 
2uqpgohn80

I think this one must be critical

3lvis:
Fixes #359


charbelhass:
@3lvis not understanding why it failed. Is it on master branch now?

3lvis:
@charbelhass it's failing because of some swift tests

charbelhass:
@3lvis ah ok,
thanks anyway i will wait to be merged with 'master'

charbelhass:
@3lvis can u tell me when it is merged with 'master' please
thank you :D

Source/FORMData.m
```diff
@@ -700,6 +711,12 @@ - (NSArray *)showTargets:(NSArray *)targets {
                 if (section) {
                     [self sectionWithID:target.targetID completion:^(FORMSection *section, NSArray *indexPaths) {
                         if (section) {
+                            for (FORMField *field in section.fields) {
```
vadymmarkov: @3lvis There is code duplication with Line 684. Maybe could be refactored by method or block?

3lvis: Will do!

3lvis: Done => https://github.com/hyperoslo/Form/commit/943bd8cdefef016842707d5dfc4778e710d46cbd

PR #382, opened 2015-04-22
[WIP] Improve/targets patch 1

3lvis:
I was playing with moving the value to live in the `FORMFieldBase` but got trapped in hell.


vadymmarkov:
@3lvis At first I've tried to move `value` to `FORMFieldBase` also, but got issues...

3lvis:
Will give it a try tomorrow, you've done a great work!

PR #398, opened 2015-05-11
Update demo to use `use_frameworks!`

zenangst:


3lvis:
This actually doesn't change anything

PR #401, opened 2015-05-11
Fix/field empty value

vadymmarkov:
When value is an empty string it causes issue with validator where we have condition `if (!fieldValue)`. Imagine the situation when we have not required email field, type something, then clear. Then field was invalid.


vadymmarkov:
screen shot 2015-05-11 at 15 40 14

PR #402, opened 2015-05-11
Don't use `self`

zenangst:


zenangst:
screen shot 2015-05-11 at 17 31 13

PR #405, opened 2015-05-12
Set timezone to GMT before running time tests

jeffleeismyhero:
Currently, date/time based tests will fail if they are run on machines in certain timezones (e.g., CST).

In order to provide a predictable string to NSDate conversion, this change sets the timezone to GMT for the purpose of testing.


zenangst:
Thanks @jeffleeismyhero, cheers :beers:

PR #417, opened 2015-05-26
Fix reset positions for sections and fields

vadymmarkov:


zenangst:
@3lvis mind having a look?

PR #421, opened 2015-06-02
Add Counter Type field with controls

jeffleeismyhero:
The added field type provides a field of type "count." This field includes two buttons to increase and decrease the fields value (by one) with each press.

screen shot 2015-06-02 at 3 37 06 pm

The field and controls allow quick and easy value adjustment without the need for a keyboard. In other respects, the field behaves like a "number" field.


jeffleeismyhero:
Thanks!

PR #437, opened 2015-06-30
Improve/reload with dictionary

zenangst:
`reloadWithDictionary:` was setting the literal value to the `FORMField` value.
This PR should fix that by looping through the `values` on the `FORMField` and use the one that matches. If it doesn't match anything in `.values`, it defaults it to `nil`.


zenangst:
screen shot 2015-06-30 at 14 51 37

vadymmarkov:
@zenangst :+1:

PR #438, opened 2015-07-08
Add unique cell identifier for Count Field.

jeffleeismyhero:
This resolves an issue where non-count cells would show the (-) and (+) views in the cell.


zenangst:
@jeffleeismyhero :beers:

Issue #445, opened 2015-08-26
Additional empty popover cells when opening popover for second time.


richardoti:
Steps to reproduce:
1. Open ObjC demo
2. Open any `FORMSelectFieldCell`, i.e. "Base Salary" or "Contract type" - Everything is correct
3. Tap on any other part of the screen to dismiss popover and open it again - observe additional empty cells and tableview growth in size.

See the screenshots attached.
Before / After
![screen shot 2015-08-26 at 23 34 01](https://cloud.githubusercontent.com/assets/8013017/9506917/3ca4e72c-4c4b-11e5-8c48-49b1c54b39d7.png)
![screen shot 2015-08-26 at 23 28 46](https://cloud.githubusercontent.com/assets/8013017/9506916/3ca4c59e-4c4b-11e5-9226-b8c5b348b2d2.png)

![screen shot 2015-08-26 at 23 28 17](https://cloud.githubusercontent.com/assets/8013017/9506922/43b9ae26-4c4b-11e5-8d4a-ce2a8092b325.png)
![screen shot 2015-08-26 at 23 28 09](https://cloud.githubusercontent.com/assets/8013017/9506921/43b29e7e-4c4b-11e5-81e7-597adec5411c.png)

3lvis:
Fixed on #446

PR #446, opened 2015-08-26
Correct size calculations

richardoti:
Does not solve #445 completely, but makes it less noticeable:
screen shot 2015-08-27 at 00 05 01 screen shot 2015-08-27 at 00 04 57


3lvis:
@richardoti great catch! :clap:

PR #456, opened 2015-09-04
Feature: localization support

zentraedi:
- Add support for NSLocalizedString(key, comment) for groups, titles,
info, field values and dynamic sections.

screen shot 2015-09-04 at 11 13 40


3lvis:
Hi Hector,

Great initiative, ideally we should just reuse the `title` key instead of adding a new one :)

Could you make that change?

zentraedi:
Sure, i’ll revert back the changes to the json file.

-- 
Héctor Cen
Dimsum C.A.
http://dimsum-apps.com

Barquisimeto, Venezuela - Latam
iOS Developer

Móvil: +58 4245595985

On 4September, 2015 at 11:46:26, Elvis Nuñez (notifications@github.com) wrote:

Hi Hector,

Great initiative, ideally we should just reuse the title key instead of adding a new one :)

Could you make that change?

3lvis:
@zentraedi you're the best!

Issue #454, opened 2015-09-03
Group header not visible


zentraedi:
Hi fellows. I just came across this excellent library for a project i need to develop.

I'm using Storyboards for my project and i tried setting up everything in different ways (demo way included) but my forms won't display the group header. I have two groups.

![screen shot 2015-09-03 at 17 20 18](https://cloud.githubusercontent.com/assets/1830701/9671900/1b05aa82-5260-11e5-995d-b8a0f43424d9.png)

This is the JSON file: https://gist.github.com/zentraedi/ff029254549ff968e55f

Additionally:

- Is it possible to use placeholders for fields?
- Is it possible to create buttons with images?
- Is it possible to specify size for dynamic sections?

Thanks in advance

jeffleeismyhero:
What happens if you add a second group?

jeffleeismyhero:
Also, when you say "placeholder for fields" are you referring to something like an empty "cell" to adjust the positioning of other fields?

jeffleeismyhero:
I put your JSON into the Basic-ObjC demo and everything worked without issue. It could be a similar issue to this: https://github.com/hyperoslo/Form/issues/444

richardoti:
@zentraedi Hi! I may look at this tomorrow. Seems that #444 might be related, asa @jeffleeismyhero  has pointed.

I will add storyboard demo project as well, so it would be possible for storyboard users to have an "official" reference.

zentraedi:
There are two groups, the second one starts with the Twitter and Facebook buttons.

I meant placeholder label for fields, like the default for UITextField provided in iOS. Could be useful if anyone prefers placeholder labels instead of the label on top of the field.

zentraedi:
Thanks @jeffleeismyhero @richardoti I'll check that out.

zentraedi:
It worked, i managed to set the layout from the presenting view controller.

Many thanks.

richardoti:
@zentraedi which way?Could you provide, how you solved this problem?

zentraedi:
Sure!

I have a simple Table View Controller embedded in a Navigation View Controller, a Bar Button Item provides a modal segue to a view controller for the form, i just set up everything in prepareForSegue, like this:

```
- (void)prepareForSegue:(UIStoryboardSegue *)segue sender:(id)__attribute__((unused))sender
{
    if ([segue.identifier isEqualToString:@"toNewEventViewController"]) {
[...]

Issue #444, opened 2015-08-26
Unaligned margins


PhilipBale:
Hi,

Seems to give: http://puu.sh/jOQKV/8cc7766f9b.png by default, using 100, 100, 50/50, 33/33/33 widths.  Should this be the behavior?

3lvis:
Hi Philip,

This doesn't seem right, would you mind providing an example of your JSON file so I can reproduce the issue? 

:)

PhilipBale:
Sure!  Here it is 
{
   "groups":[
     {
      "id":"group-id",
      "title":"Group title",
      "sections":[
         {
            "id":"section-0",
            "fields":[
               {
                  "id":"title",
                  "title":"Title*",
                  "type":"text",
                  "size":{
                     "width":100
                  }
               },
               {
                  "id":"location",
                  "title":"Location*",
                  "type":"text",
                  "size":{
                     "width":100
                  }
               },
               {
                  "id":"date",
                  "title":"Date*",
                  "type":"date",
                  "size":{
                     "width":50
                  }
               },
               {
                  "id":"attire",
                  "title":"Attire*",
                  "type":"select",
                  "size":{
                     "width":50
                  },
                  "values":[
                     {
                        "id":0,
                        "title":"Casual",
                        "default":true
                     },
                     {
                        "id":1,
                        "title":"Formal"
                     }
                  ]
               },
               {
                  "id":"size",
                  "title":"Size*",
                  "type":"text",
                  "size":{
                     "width":33
                  }
               },
               {
                  "id":"min_age",
                  "title":"Min age",
                  "type":"number",
                  "size":{
                     "width":33
                  }
               },
               {
                  "id":"max_age",
                  "title":"Max age",
                  "type":"number",
                  "size":{
                     "width":33
                  }
               }
            ]
         }
      ]
   }
  ]
}

jeffleeismyhero:
When using the provided JSON below in the latest commit of Form, I get the following form:

![screenshot 2015-08-26 08 18 33](https://cloud.githubusercontent.com/assets/43980/9494442/77056782-4bcb-11e5-9b6d-07d4c0061219.png)

What version of Form are you using?

**Provided JSON**
```json
{  
    "groups":[  
        {  
            "id":"group-id",
            "title":"Group title",
            "sections":[  
                {  
                    "id":"section-0",
                    "fields":[  
                        {  
                            "id":"title",
                            "title":"Title",
                            "type":"text",
                            "size":{  
                                "width":100
[...]

PR #457, opened 2015-09-04
Storyboard demo

richardoti:
I've added storyboard demo to the project as we've seen few issues: #454 and #444 where there were questions about how to use Form with storyboards.

I've used iPhone as a device and used JSON from #444 as default JSON does not look good on smaller screens.

Here is the result:
ios simulator screen shot 05 sep 2015 01 24 40


3lvis:
@richardoti you might want to check your CocoaPods version, you're using a beta version.

richardoti:
@3lvis I know. The situation is under control :+1:  , but it's good that you noticed.

Issue #419, opened 2015-05-27
Crash in FORMFieldValue's identifierIsEqualTo:


3lvis:
Stacktrace:
```
2015-05-27 14:22:59.855 [12230:66511] -[FORMFieldValue objCType]: unrecognized selector sent to instance 0x7faae9ec0700
```

3lvis:
Closing for lack of steps to reproduce.

PR #461, opened 2015-09-08
Refactor FORMFieldValuesTableViewHeader

zenangst:
## Refactor initializer
FORMFieldValuesTableViewHeader is a UITableViewHeaderFooterView so
`initWithFrame` was never called.

The side effect of this was that the subview were never added to the
view.

## Use title instead of rawValue
This was causing issues when running it with a new version of Xcode.

```
[__NSCFNumber isEqualToString:]: unrecognized selector sent to instance
0x7bb73b10
```

The initial value that was assigned to `.text` was a NSNumber and when
it tried to assign it to a string it couldn’t compare the two.

### DIFF
--- a/Demos/Basic-ObjC/Podfile.lock
+++ b/Demos/Basic-ObjC/Podfile.lock
@@ -1,5 +1,5 @@
 PODS:
-  - Form (3.5.0):
+  - Form (3.6.0):
     - Hex (~> 1.1.1)
     - HYP8601 (~> 0.7.2)
     - HYPImagePicker (~> 0.3)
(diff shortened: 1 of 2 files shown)


richardoti:
I may guess, this is related to bridging with Swift, as `initWithReuseIdentifier` is a designated initializer.

zenangst:
@richardoti well not really, in the case where I was using it, it doesn't use Swift.

richardoti:
@zenangst I don't mean the issue is due to usage with Swift, but in general. I remember, Apple did some bridging with ObjC, so it may be related.

3lvis:
@zenangst Can we keep the old `initWithFrame` too for backwards compatibility?

richardoti:
Is this issue relevant? #419

zenangst:
@3lvis well, it was never called in the code so I felt that it was unnecessary to keep it around.

https://github.com/hyperoslo/Form/blob/master/Source/Cells/Popover/FORMFieldValuesTableViewController.m#L21

richardoti:
Tested on Basic ObjC demo, works OK.

3lvis:
@zenangst makes sense :)

3lvis:
@richardoti It's great to get your feedback, you are super helpful, but could let merges to either @zenangst or me? :heart:  

https://github.com/hyperoslo/playbook/blob/master/GIT_AND_GITHUB.md#who-merges-the-pull-request

richardoti:
@3lvis OK. @zenangst asked me to merge, so I think, this time it doesn't count. :sunglasses:

PR #462, opened 2015-09-08
Only show info label if it has content

zenangst:
When we introduced the localization of `.info` strings, we forgot that we depend on `.info` being empty or not to decide if we should show or the tooltip. This pull request fixes that issue by looking for `.length` of `.info` before displaying it.

Ref; https://github.com/hyperoslo/Form/blob/8f5355368a4d628fa5bcbe517d14798a56b61e8d/Source/Models/FORMField.m#L38-38

### DIFF
--- a/Source/Cells/Text/FORMTextFieldCell.m
+++ b/Source/Cells/Text/FORMTextFieldCell.m
@@ -273,7 +273,7 @@ static NSString * const FORMTooltipBackgroundColorKey = @"tooltip_background_col
 }
 
 - (void)showTooltip {
-    if (self.field.info && [self.field.info length] > 0 && self.showTooltips) {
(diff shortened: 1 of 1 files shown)


Source/Cells/Text/FORMTextFieldCell.m
```diff
@@ -273,7 +273,7 @@ - (CGRect)textFieldFrame {
 }
 
 - (void)showTooltip {
-    if (self.field.info && self.showTooltips) {
+    if (self.field.info && [self.field.info length] > 0 && self.showTooltips) {
```
richardoti: Looks good to me.
You may want to use `self.field.info.length` here as well to avoid `[]`

PR #465, opened 2015-09-24
[WIP] Improvement: payment demo

3lvis:
A demo on how to make a payment form. Wow, this was way too easy! :clap: 

payment

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+osx_image: xcode7
 language: objective-c
 cache: cocoapods
 podfile: Tests/Podfile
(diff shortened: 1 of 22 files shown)


3lvis:
simulator screen shot 24 sep 2015 14 49 16

jeffleeismyhero:
I've been looking to create support for label fields. This is a great example of where they would be very useful. Where do you display the amount someone is paying? You could use a disabled form field but you're a bit limited with sizing and such.

Secondly, the iPad doesn't have a nice number pad. That makes number entry more cumbersome than it needs to be. For one of my applications, I have a PIN pad that could probably be pushed into a popover style view.

keypad

Thoughts?

3lvis:
@jeffleeismyhero Ideally those additional fields should live in separate repos or in the apps that implement them. We should start removing code from Form instead of adding it if we want to continue supporting it in the future.

More more into a plug-and-play component instead of a component that has all the things.

jeffleeismyhero:
I agree with you there. I believe the barrier to entry is that there are no fully external plug-and-play components that pave the way and set the standards others should follow.

I would love to see an example of a fully built custom component that could be included in a project as a Pod.

How can we get there?

3lvis:
@jeffleeismyhero I think support for external components isn't implemented at the moment, but the plan is to have that before adding more custom fields into Form.

PR #466, opened 2015-09-25
Improve: move input validator and formatter dependencies out of form

3lvis:
This dependencies should be able to be used without needing Form. This is the first step of abstracting the super powerful texfield for better reusability.


3lvis:
Any input on this @zenangst ?

Issue #486, opened 2015-11-17
appDelegate and initCoder


ehirebrandon:
Everything seems to be have to work through app delegate what if I want to use View controller. I've try so many ways, I can't subclass without an error and have it running.

3lvis:
@ehirebrandon could you share the error with us? This should work. Are you using storyboards? There's a sample project on how to do this.

https://github.com/hyperoslo/Form/tree/master/iPhone-Storyboard

ehirebrandon:
Hi,

Yes I am trying to use storyboard. What I try to do is subclass (let's say
like a sampleViewController that subclass from FORMViewController). Then in
my view controller I try to present the an instance of the
sampleViewController however it does not work. It seems like the only way
is to go through app delegate to present the form, which works. However I
have multiple view controller that is in an array that let me slides from
one to another. However Json error/ and other random errors come up. I'm
currently using swift and I can't understand FORMViewcontroller etc, so I
only hope to subclass from it and everything work. Can you please update
with an example that have view controller that uses it without the app
delegate. The sample for iphone-storyboard is also insufficient, because it
does not provide a view controller that present it the way most other apps
do. It's quite unconventional for anyone to just use these forms as one
single view. Can you assist?

Thank you,

brandon

3lvis:
@ehirebrandon Here it is the demo, there's an issue with the layout's delegate, I'm not experienced on storyboards, hopefully you'll be able to fix it :)

https://github.com/hyperoslo/Form/pull/487

3lvis:
@ehirebrandon https://github.com/hyperoslo/Form/tree/master/ModalForm 

Here's how you do it. https://github.com/hyperoslo/Form/blob/master/ModalForm/ViewController.swift#L6-L7

You don't need to hook up anything on Interface Builder.

Make sure you apply the default style https://github.com/hyperoslo/Form/blob/master/ModalForm/AppDelegate.swift#L11

PR #487, opened 2015-11-17
Demo: modal storyboard

3lvis:
For some reason the layout is not holding the dataSource, so it crashes. Everything else works just fine for https://github.com/hyperoslo/Form/issues/486

screen shot 2015-11-17 at 22 00 19


3lvis:
Any ideas on what's going on?

3lvis:
FORMViewController relays heavily on it's initializer. Using `initWithCoder` breaks the entire API. I suggest we just drop `initWithCoder` support and let people present their Forms programmatically.

ehirebrandon:
please!

richardoti:
@3lvis I remember having troubles with this part as well. In fact, there is little need to initialize Form completely in storyboard, at least from my point of view.

This framework is made to work in code, as the content is being generated using a JSON and can vary a lot. And setting `FORMLayout` in storyboard doesn't make it  easier to use. The level of the framework is not for beginners, so they should know how to initialize it proper way.

The only real need to use FORM Form with storyboards is to to be able to see general app flow and to configure navigation bar buttons with ease.

So I agree, it's totally fine to leave it as it is :smile:

3lvis:
Yeah, using Form without storyboards is simple enough.

PR #488, opened 2015-11-20
Improve setting selected text range

zenangst:
**Before**
jumpy

**After**
no-jumpy


zenangst:
screen shot 2015-11-20 at 12 26 38

vadymmarkov:
@zenangst :+1:

PR #496, opened 2016-01-01
Section separators should be hidden by default

3lvis:
Otherwise they show up when creating custom fields

screen shot 2016-01-01 at 18 27 26


3lvis:
screen shot 2016-01-01 at 18 32 00

PR #497, opened 2016-01-01
Remove excess of white space in CustomField demo

3lvis:
# Before 

screen shot 2016-01-01 at 18 32 57

# After

screen shot 2016-01-01 at 18 33 21

### DIFF
--- a/CustomField/Resources/forms.json
+++ b/CustomField/Resources/forms.json
@@ -60,7 +60,7 @@
               "type":"subtitle",
               "size":{
                 "width":100,
-                "height":1
+                "height":0.7
               }
             },
             {
--- a/CustomField/SubtitleField.swift
+++ b/CustomField/SubtitleField.swift
@@ -7,11 +7,8 @@ class SubtitleField: FORMBaseFieldCell {
     lazy var textView: UITextView = {
         let horizontalMargin = 10.0
         let width = Double(self.frame.width) - horizontalMargin * 2.0
-
-        let topMargin = 30.0
-        let height = Double(self.frame.height) - topMargin
-
-        var frame = CGRect(x: horizontalMargin, y: topMargin - 20, width: width, height: height)
+        let height = Double(self.frame.height)
+        var frame = CGRect(x: horizontalMargin, y: 0, width: width, height: height)
         let view = UITextView(frame: frame)
 
         view.font = UIFont(name: "AvenirNext-Regular", size: 15.0)!


RyanKim01:
One quick question. If textfield does not have a title "Start date" then it looks like there's more blank between a sentence and a text field. Is there a way to completely take out the space of "start date"?

3lvis:
Hi, fields have titles, we haven't added support for fields without titles and doesn't seems like it's something we'll do. I think fields need titles in order to make users understand what's inside them.

RyanKim01:
@3lvis , I have many questions with many sentences. So it does not fit into Form's default state which only shows one line of a question. Thus, I was utilizing custom field to put multiple lines of questions and putting textfield without a question below. But this way created an awkward white gap between the custom field and the textfield as I mentioned. So I was curious about a way to accomplish this without any gap.

3lvis:
I guess what you really want is that fields support multiline titles. I'll add an issue for that :)

RyanKim01:
@3lvis YES! I figured out the more I try other ways the more the layout becomes weird :/  Thank you for adding the issue!

3lvis:
Btw, do you have a real world use case of this? Would it be better to have title and subtitle instead of multiline titles?

RyanKim01:
@3lvis , Yes. I'm basically digitizing a form into iOS app right now and my form contains a lot of multiline questions and this drives me nuts right now haha. I think putting multiline options in title, popover cells, and textfield(select type) would be really good. Basically, if it is possible to put multiline options in anywhere else, that would be really awesome. If that is too much, I think multiline title is a good start. 

Moreover, I've encountered few times that I needed title and subtitle options. But I somehow got around with it by using existing options like groups, sections etc. It would be very delightful to see title & sub-title though :)

3lvis:
@RyanKim01 But do you have a real life use case example? For example you provided this:

7c4c32ba-a4cd-11e5-8e5a-909566c1ddaa

Here you don't need multiline field titles. Do you have an example where you need that?

RyanKim01:
@3lvis oh my bad. I understood it in a wrong way. Here is an example.
screen shot 2016-01-01 at 10 20 44 pm

As you can seem a question has 4 sentences and simple yes/no answer. Select type can be used  for this kind of a question, but title should be multilines.

3lvis:
This makes a lot of sense.  Thanks!

RyanKim01:
@3lvis  Thanks for adding it :)

3lvis:
@RyanKim01 It's up for grabs, feel free to work on it. I'm probably not going to be able to look into it for now.

PR #503, opened 2016-01-11
Made cell height in popover dynamic

RyanKim01:
Use case for this is when users have many sentences in one cell in "select" type of questions. The change I made will make the cell height in popover dynamic so all sentences can be shown.


3lvis:
Nice!

3lvis:
Available in 3.9.0 

https://github.com/hyperoslo/Form/releases/tag/3.9.0

PR #506, opened 2016-01-21
Fix demo crash

richardoti:
All ObjC demos crashed on startup because of not enabled flag "Embedded Content Contains Swift Code". I fixed these issues by enabling flag on all demo targets.

SO question:
https://stackoverflow.com/questions/26024100/dyld-library-not-loaded-rpath-libswiftcore-dylib


3lvis:
Thanks!

zenangst:
@richardoti 
unknown-5

PR #513, opened 2016-03-09
Don't reset field positions when hiding fields

zenangst:
We were experiencing issues with showing hidden fields. They ended up in the wrong position because Form mutates the state after the field becomes hidden, so if you hide two fields in a sequence and try to show them again, the end up being displayed in the wrong order.

This PR removes the reset of positions when hiding fields, tests pass and our internal testing has shown positive results that this won’t blow up.

chkb0fsuwaizqy5

### DIFF
--- a/Source/FORMData.m
+++ b/Source/FORMData.m
@@ -798,10 +798,6 @@ includingHiddenFields:(BOOL)includingHiddenFields
                        }];
     }
 
-    [resetSections enumerateObjectsUsingBlock:^(FORMSection *section, BOOL *stop) {
-        [section resetFieldPositions];
-    }];
-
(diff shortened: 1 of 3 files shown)


RamonGilabert:
:scream: @vadymmarkov :skull:

vadymmarkov:
@RamonGilabert Tests passed, am I right? :grimacing:

zenangst:
@vadymmarkov 

screen shot 2016-03-09 at 13 06 32

RamonGilabert:
Well, if tests pass then I guess we could just...

giphy

3lvis:
:scream:

PR #516, opened 2016-04-22
Remove Hex dependency

3lvis:
Remove requirement so Form can be used without `use_frameworks!`


3lvis:
Decided to not merge this since it breaks Swift projects

PR #528, opened 2016-06-23
Use field type as the default input type

3lvis:
Fixes #528


3lvis:
screen shot 2016-06-23 at 11 08 34 pm

PR #542, opened 2016-10-06
Add support for Accessibility Labels on text field

jeffleeismyhero:
This adds support for setting the `accessibilityLabel` attribute on UITextField through JSON.

**Attribute Name:** `accessibility_label`
**Applies To:** Text type fields

**Usage**
```json
{
  "id":"first_name",
  "title":"First name",
  "accessibility_label":"First Name Accessibility Label",
  "type":"text",
  "size":{
    "width":30,
    "height":1
  }
}
```

The field's `title` attribute will be used for the accessibility label in the event one is not provided.

### DIFF
--- a/Source/Cells/Popover/FORMSelectFieldCell.m
+++ b/Source/Cells/Popover/FORMSelectFieldCell.m
@@ -62,6 +62,13 @@ static const NSInteger FORMSelectMaxItemCount = 6;
     } else {
         self.fieldValueLabel.text = nil;
     }
+    
+    if ([field.accessibilityLabel length] > 0) {
+        self.fieldValueLabel.accessibilityLabel = field.accessibilityLabel;
+    } else {
+        self.fieldValueLabel.accessibilityLabel = field.title;
+    }
+    self.fieldValueLabel.accessibilityValue = self.fieldValueLabel.text;
 }
 
 #pragma mark - FORMPopoverFormFieldCell
--- a/Basic-ObjC/Resources/forms.json
+++ b/Basic-ObjC/Resources/forms.json
@@ -10,6 +10,7 @@
             {
               "id":"first_name",
               "title":"First name",
+              "accessibility_label":"First Name Accessibility Label",
               "info":"Fornavn",
               "type":"name",
               "disabled":true,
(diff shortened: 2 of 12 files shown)


3lvis:
This is a great feature, @jeffleeismyhero! 

Would it be possible to add support to the other fields as well? I'm thinking on the multiselects and so on.

jeffleeismyhero:
Certainly. I've added support for date, time, and select fields.

The selected value is also set to `accessibilityValue` for those fields.

Source/Cells/Popover/FORMSelectFieldCell.m
```diff
@@ -62,6 +62,13 @@ - (void)updateWithField:(FORMField *)field {
     } else {
         self.fieldValueLabel.text = nil;
     }
+    
+    if ([field.accessibilityLabel length] > 0) {
+        self.fieldValueLabel.accessibilityLabel = field.accessibilityLabel;
+    } else {
+        self.fieldValueLabel.accessibilityLabel = field.title;
+    }
+    self.fieldValueLabel.accessibilityValue = self.fieldValueLabel.text;
```
3lvis: Looks like this will overwrite whatever happened before this.

jeffleeismyhero: The if statement is setting the accessibility **label** (`accessibilityLabel`), the line outside of that is setting the accessibility **value** (`accessibilityValue`).

If the `accessibility_label` JSON value is blank or not defined, it will use the `field.title` value otherwise it will use the defined label. After that, it will set the `accessibilityValue` property to the selected value as it is printed in the field.

**Assuming:**
`accessibility_label` = Visit time
`self.fieldValueLabel.text` = 6:00 PM

**The results is this voiceover audio:**
`Visit time six o'clock pm`

3lvis: oHHHHHHHH

3lvis:
Fantastic work @jeffleeismyhero!

3lvis:
@jeffleeismyhero could you document this in the README, please?

PR #549, opened 2016-10-22
Clean up

3lvis:
- Simplify Travis
- Update HYPMathParser to finally remove that annoying warning

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 osx_image: xcode8
 language: objective-c
 cache: cocoapods
+before_script: pod update
 script: xcodebuild -workspace Demo.xcworkspace -scheme Tests -sdk iphonesimulator test | xcpretty -c && exit ${PIPESTATUS[0]}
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -1,4 +1,5 @@
-osx_image: xcode7.3
+osx_image: xcode8
 language: objective-c
 cache: cocoapods
-script: xcodebuild -workspace Demo.xcworkspace -scheme Tests -sdk iphonesimulator build test GCC_INSTRUMENT_PROGRAM_FLOW_ARCS=YES GCC_GENERATE_TEST_COVERAGE_FILES=YES clean test | xcpretty -c && exit ${PIPESTATUS[0]}
+before_script: pod update
```
zenangst: Wouldn't it be safer to use `pod install` instead of `pod update` ?
Depends on the `Podfile` I guess.

3lvis: You need `pod update` since `pod install` doesn't update the specs repo if needed.

vadymmarkov:
@3lvis I've managed to fix Travis in https://github.com/hyperoslo/Form/pull/551

PR #550, opened 2016-10-31
Highlight disabled text field if it is not valid

vadymmarkov:
@3lvis @zenangst 

Validation works for disabled fields as well, but they were not highlighted with `invalidColor`. This PR aims to fix that.


Source/Cells/Text/FORMTextFieldCell.m
```diff
@@ -177,6 +177,8 @@ - (void)updateFieldWithDisabled:(BOOL)disabled {
 - (void)updateWithField:(FORMField *)field {
     [super updateWithField:field];
 
+    [self validate];
```
vadymmarkov: Let me know if you guys see any implication of calling `validate` here. I's done to update `isValid` property considering the new value whenever the field is changed.

vadymmarkov:
I see https://github.com/hyperoslo/Form/pull/549 fixes Travis, so it's better to get it merged before this one I guess.

zenangst:
@vadymmarkov 🚀

zenangst:
Lets try and fix the Travis build :)

PR #479, opened 2015-10-13
Feature: popover navigation controller

richardoti:
Issue #452 
screen shot 2015-10-13 at 15 55 15
screen shot 2015-10-13 at 15 58 33

screen shot 2015-10-13 at 15 59 57

Everything is ready, the only question is how to apply styles to `FORMFieldValuesTableViewController` as it assigns colors of `Cancel`, `Title` and `Done` button styles. Works on both iPad and iPhone.

Still can't set up styles and I have no idea, why. This is the only part that is not ready:
screen shot 2015-10-14 at 11 00 36

PR #551, opened 2016-11-01
Fix/ipad popovers

vadymmarkov:
Feature introduced in https://github.com/hyperoslo/Form/pull/479 caused some issues in popover view on iPad, such us:
- Long titles were cut for the date field because of 3 buttons in the navigation bar;
- Clear button should not be displayed for every select field.

This PR reverts to old popover UI on iPad, so the feature with navigation controller will be used only on iPhone. It's more a "hotfix", so this functionality could be re-designed and changed later if needed.

### DIFF
--- a/.slather.yml
+++ b/.slather.yml
@@ -1,4 +1,0 @@
-ci_service: travis_ci
-coverage_service: coveralls
-xcodeproj: Demo.xcodeproj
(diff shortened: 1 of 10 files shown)


3lvis:
🎉

PR #552, opened 2016-11-11
Clearing a select fields causes a crash on subsequent opening

jeffleeismyhero:
When a select field is cleared using the "clear" button, subsequent access causes `- (BOOL)identifierIsEqualTo:(id)identifier` to hit the `else` condition which calls `abort()`.

This removes the `else` condition and allows the method to return `NO`.


3lvis:
Hi @jeffleeismyhero, you're a maintainer now, so you can make a branch in Form. You don't need to use your own branch anymore.

PR #544, opened 2016-10-10
Segmented Control / Switch Field

jeffleeismyhero:
This adds the functionality requested in the Feature Request Issue #536. It might also work for the functionality requested in issues #474 and #476.

**Field type:** `segment`
**Values:** This field takes values in the same format as select fields.

**Example JSON**
```json
[
  {
    "id": "group",
    "title": "Group",
    "sections": [
      {
        "id": "section",
        "fields": [
          {
            "id": "segment",
            "title": "Location",
            "type": "segment",
            "values":[
              {
                "id":"in_house",
                "title":"In-house",
                "default":true,
              },
              {
                "id":"remote",
                "title":"Remote",
              }
            ],
            "size": {
              "width": 40,
              "height": 1
            }
          }
        ]
      }
    ]
  }
]
```

Issue #536, opened 2016-09-05
Feature Request: Segmented Control / Switch Field


jeffleeismyhero:
Re: #474 and #476

A solution for booleans and an alternative for select fields can be very helpful - especially for one or two options. Both scenarios can be resolved with a UISegmentedControl based solution.

<img width="369" alt="screen shot 2016-09-05 at 5 42 45 pm" src="https://cloud.githubusercontent.com/assets/43980/18257852/39d49ed0-7390-11e6-8ca4-f4384d27e407.png">
<img width="150" alt="screen shot 2016-09-05 at 5 44 27 pm" src="https://cloud.githubusercontent.com/assets/43980/18257866/76a4288a-7390-11e6-9374-bc801f6dfdea.png">

I'm happy to tackle this next week if you'll have it.

3lvis:
![ios-7-style-smooth-toggle-buttons-with-jquery-css3-checkable](https://cloud.githubusercontent.com/assets/1088217/19250799/93cbbffa-8f3c-11e6-9e75-909bd7f0bab5.jpg)

3lvis:
Why not that one instead?

jeffleeismyhero:
Ideally, I'd like to add support for UISwitch as well. In the meantime, UISegmentedControl fields are versatile enough to handle the UISwitch use-cases AND work well with FORM's width-based sizing.

Additionally, UISegmentedControl fields can support more than yes/no based selections.

3lvis:
Makes sense. I guess we'll keep the Switch issues open and offer this as a workaround.

Issue #476, opened 2015-09-29
Toggle for Yes/No fields


3lvis:
https://github.com/hyperoslo/Form/issues/474

jeffleeismyhero:
Would this be different from a Switch? Like a button with an on/off state?

3lvis:
I think you already did this one.

Issue #474, opened 2015-09-28
Can we have toggle


annehiro:
A drop down for Yes and No doesn't seem like the best UX.

annehiro:
Or is there any way I can replace the textField with a UISwitch?

annehiro:
Ah, custom field.

3lvis:
@annehiro I'll make an issue for the toggle, great feature request!

PR #555, opened 2016-11-12
Add support for Toggle (UISwitch) fields

jeffleeismyhero:
This PR adds support for Toggle fields as requested in #474, #476 and referenced in #536 and #544.

The field supports styling for tint and background colors. You can also set the default value using: `0`, `false`, `1`, `true`

```json
{
  "id":"toggle1",
  "title":"Toggle Switch",
  "type":"toggle",
  "value":true,
  "size":{
    "width":50
  },
  "styles":{
    "tint_color":"#3DAFEB",
    "background_color":"#FFFFFF"
  }
}
``` 

screen shot 2016-11-12 at 1 31 50 pm


3lvis:
Hi @jeffleeismyhero,

This is a great addition! Damn, you're on fire!

I only have one feedback. Should be switch instead of "toggle". If you look around most implementations of "toggle" mean something different.

jeffleeismyhero:
I was originally going to use "switch" but ran into a conflict since
"switch" is a reserved word.

3lvis:
Let's just ship it.

PR #556, opened 2016-11-13
Use Switch instead of Toggle

3lvis:
Made it work!


jeffleeismyhero:
This looks good to me!

Issue #559, opened 2016-12-03
NSRangeException when scrolling to bottom of form with show target action


jeffleeismyhero:
As shown in the screen capture below, an `NSRangeException` is thrown when `processTargets` is called with a show target action after scrolling to the bottom of a form.

This crash seems to depend on hitting the view's bounce behavior in conjunction with a show target action. If I remove the show target action OR do not trigger the bounce behavior, the crash does not occur. I am using the default `forms.json` in the `Basic-ObjC` demo on master.

![form-nsrangeexception](https://cloud.githubusercontent.com/assets/43980/20862081/f8ac3e50-b966-11e6-8096-2d08892314e6.gif)

```objective-c
2016-12-03 14:29:41.313 Basic-ObjC[11419:14354078] *** Terminating app due to uncaught exception 'NSRangeException', reason: '*** -[__NSArrayM objectAtIndex:]: index 7 beyond bounds [0 .. 6]'
*** First throw call stack:
(
	0   CoreFoundation                      0x000000010cc4a34b __exceptionPreprocess + 171
	1   libobjc.A.dylib                     0x000000010c6ab21e objc_exception_throw + 48
	2   CoreFoundation                      0x000000010cb7bf1b -[__NSArrayM objectAtIndex:] + 203
	3   UIKit                               0x000000010e16e37d -[_UIFlowLayoutSection frameForItemAtIndexPath:] + 155
	4   UIKit                               0x000000010e109eb5 -[UICollectionViewFlowLayout _frameForItem:inSection:usingData:] + 159
	5   UIKit                               0x000000010e103a72 -[UICollectionViewFlowLayout layoutAttributesForItemAtIndexPath:usingData:] + 561
	6   UIKit                               0x000000010e103d23 -[UICollectionViewFlowLayout layoutAttributesForItemAtIndexPath:] + 406
	7   Form                                0x000000010bf8ea1c -[FORMLayout layoutAttributesForDecorationViewOfKind:atIndexPath:] + 796
	8   Form                                0x000000010bf8f876 -[FORMLayout layoutAttributesForElementsInRect:] + 1526
	9   UIKit                               0x000000010e124be0 __45-[UICollectionViewData validateLayoutInRect:]_block_invoke + 241
	10  UIKit                               0x000000010e1244e3 -[UICollectionViewData validateLayoutInRect:] + 2984
[...]

PR #560, opened 2016-12-03
Disable collectionView bounces to prevent NSRangeException

jeffleeismyhero:
This resolves the crash presented in issue #559. It eliminates the crash by disabling bounces on the FORMViewController.

Note: This crash only occurs on iOS 10 so the underlying problem may be outside of Form.

### DIFF
--- a/Source/FORMViewController.m
+++ b/Source/FORMViewController.m
@@ -41,6 +41,8 @@
         [self.collectionView numberOfSections];
     }
 
+    self.collectionView.bounces = NO;
+    
     [self hyp_addKeyboardToolbarObservers];
 
     return self;


3lvis:
Hi @jeffleeismyhero,

Thanks! Could you add a comment on top of the added code stating why this is needed? So we make sure it's not removed.

3lvis:
I gave it a try https://github.com/hyperoslo/Form/pull/561

3lvis:
Since the crash is still present in #561, merging this.

3lvis:
I don't really like that it doesn't bounce, though :(

Issue #429, opened 2015-06-19
Sort Order


jeffreyjackson:
What causes targets to show up at the top of a section when they are shown?  Ideally, I'd like all of my targets to maintain order with how they are in the JSON file.  Am I missing something?

3lvis:
Hi Jeffrey,

They should keep the original location, do you have an example JSON to debug what's going on?

Best,

jeffreyjackson:
this is the forms.json I'm using in basic objc project
```
{
  "groups": [
    {
      "id": "group-dialysis",
      "title": "Dialysis",
      "sections": [
        {
          "id": "section-dialysis-dialysis",
          "title": "Dialysis",
          "fields": [
            {
              "id": "di_dialysis",
              "title": "Dialysis?",
              "type": "select",
              "values": [
                {
                  "id": "0",
                  "title": "N/A",
                  "targets": [
                    {
                      "id": "di_dialysis_type",
                      "action": "hide",
                      "type": "field"
                    },
                    {
                      "id": "di_schedule",
                      "action": "hide",
                      "type": "field"
                    },
                    {
                      "id": "di_access",
                      "action": "hide",
                      "type": "field"
                    },
                    {
                      "id": "di_transfer_type",
                      "action": "hide",
                      "type": "field"
                    },
                    {
                      "id": "di_comments",
                      "action": "hide",
                      "type": "field"
                    },
                    {
                      "id": "di_transfer_type_details",
                      "action": "hide",
                      "type": "field"
                    },
                    {
                      "id": "di_schedule_specify",
                      "action": "hide",
                      "type": "field"
                    }
                  ]
                },
                {
                  "id": "1",
                  "title": "Yes",
                  "targets": [
                    {
                      "id": "di_dialysis_type",
                      "action": "show",
                      "type": "field"
                    },
                    {
                      "id": "di_schedule",
                      "action": "show",
                      "type": "field"
                    },
                    {
                      "id": "di_access",
                      "action": "show",
                      "type": "field"
                    },
                    {
                      "id": "di_transfer_type",
                      "action": "show",
                      "type": "field"
                    },
                    {
                      "id": "di_comments",
                      "action": "show",
                      "type": "field"
                    }
                  ]
                }
              ]
            },
            {
              "id": "di_dialysis_type",
              "title": "Type",
              "type": "select",
              "hidden": "true",
              "values": [
                {
                  "id": "0",
                  "title": "ACUTE"
                },
                {
                  "id": "1",
[...]

PR #571, opened 2017-02-20
Hidden field order

jeffleeismyhero:
This PR resolves an issue identified in #429. When fields are initially hidden and subsequently shown, they appear at the top of the section rather than their JSON defined arrangement.

This solution seems inelegant, but works -- which is why I'm requesting review.


3lvis:
Hi @jeffleeismyhero,

Would you mind adding a unit test so we make sure this doesn't break in the future?

Thanks!

3lvis:
We can add the unit test later ;)

PR #588, opened 2018-04-30
Fix bug where date picker keeps reducing height when showing again

AnthoPakPak:
In the demo project, if you click on a date field several times, the picker height will keep decreasing, and finally become totally hidden.
This line fixes this.


3lvis:
Thanks!

3lvis:
Thanks @AnthoPakPak!

PR #594, opened 2018-08-06
Support multi_select!

chy89310:
1. Add FORMFieldTypeMultiSelect using FORMSelectFieldCell.
2. Use NSArray to initialise and retrieve multiple select values.
3. Support multiple select in FORMFieldValuesTableViewController.
4. Support validations.
5. Support targets.

Thanks for Marcus who originally come up with this solution!

aug-06-2018 12-51-20


3lvis:
#yolo merge :)

Thanks for this!

